PR #27, opened 2019-02-25
Added Script that can connect a bluetooth Xbox controller to the RPi

zeinyoussef:


src/bluetooth.py
```diff
@@ -0,0 +1,30 @@
...
+            print(bluetooth.lookup_name(i))
+            if targetName == bluetooth.lookup_name(i):
+                targetAddress = i
+
+        time.sleep(1)
+
+    print(targetAddress)
+
+
+    child = pexpect.spawn('bluetoothctl')
```
johnshamoon: This is exactly what I told you not to do. This is running the exact same commands that we are doing currently in bash. This just adds another layer of slow, unnecessary Python complexity. If this is the only way to do it, maybe this should just be done in bash.

src/bluetooth.py
```diff
@@ -0,0 +1,30 @@
...
+    child.sendline('connect ' + targetAddress)
+    child.sendline('exit')
+    child.interact()
+
+
+if __name__ == '__main__':
+    targetName = "Xbox Wireless Controller"
```
johnshamoon: There can be plenty of devices with the name "Xbox Wireless Controller." It's a lot more reliable and precise to use the Bluetooth address.

johnshamoon: Also you could just pass this as a parameter. Probably don't need a variable for it.

src/bluetooth.py
```diff
@@ -0,0 +1,30 @@
+import bluetooth
```
johnshamoon: Where is this bluetooth coming from? If you require special packages, you should create a requirements.txt to let us know what packages we need to install. Same with pexpect if that's not built in.

johnshamoon:
[Work on a script to connect Bluetooth to Python](https://trello.com/c/kTUbMn2L/85-work-on-a-script-to-connect-bluetooth-to-python)

src/bluetooth.py
```diff
@@ -0,0 +1,30 @@
...
+import pexpect
+import time
+
+def connect(targetName):
+    targetAddress = None
+
+    while targetAddress == None:
```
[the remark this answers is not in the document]
zeinyoussef: I can change that no problem, I just didn't know what the address for our controller was when I wrote this.

src/bluetooth.py
```diff
@@ -0,0 +1,30 @@
...
+        nearby_devices = bluetooth.discover_devices()
+        print(len(nearby_devices))
+        for i in nearby_devices:
+            print(bluetooth.lookup_name(i))
+            if targetName == bluetooth.lookup_name(i):
+                targetAddress = i
+
+        time.sleep(1)
```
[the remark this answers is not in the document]
zeinyoussef: Noted, I can take that out. I just slept for testing.

src/bluetooth.py
zeinyoussef: Yes, this is the only way I found that it could work. The L2CAP required a client and a server and I found another package called gatt but it would not work because it requires us to scan for an adapter.

src/bluetooth.py
zeinyoussef: I agree, I can take that out once I know the address for our controller, but this just made it easier to test at home with my controller.

src/bluetooth.py
zeinyoussef: I'll get on that right now.

src/bluetooth.py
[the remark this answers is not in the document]
johnshamoon: 5C:BA:37:26:6D:9A Xbox Wireless Controller

src/bluetooth.py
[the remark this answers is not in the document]
johnshamoon: I don't think we need to scan anyway. If we're paired to a device, we should just be able to connect. We don't need to put the Pi in discovery mode again because that searches for devices that we haven't paired with. We definitely don't need to be printing and that should be taken out before pushing. It's extremely useful for development, but it's not helpful for the actual software and will slow it down a lot. Printing is a blocking operation so taking it out should make this faster.

src/bluetooth.py
johnshamoon: The specific Bluetooth packet protocol isn't what I'm concerned about. It doesn't matter if you're using L2CAP or GATT (you're scanning anyway) or RFCOMM or whatever. What matters is that you're still doing exactly what we're doing manually. 

> ```
>     child = pexpect.spawn('bluetoothctl')
>     child.sendline('connect ' + targetAddress)
>     child.sendline('exit')
>     child.interact()```

All this code is doing is literally sending "bluetoothctl" to the terminal and then piping "connect $ADDR" to it. I'm pretty sure this could be done with just one command in bash:
 ```
echo -e connect `5C:BA:37:26:6D:9A\nquit' | bluetoothctl
```

I don't think this Python solution is better than that one liner in bash. The Python is definitely slower. I gave you this task to see if there's a better way to do this. This task is crucial to the project.

File: src/bluetooth.py [hunk 6]
```diff
@@ -0,0 +1,25 @@
+import pexpect
+import time
+
+import pexpect
```
johnshamoon: imported pexpect and time twice.

src/bluetooth.py
```diff
@@ -0,0 +1,25 @@
...
+import time
+
+def connect():
```
johnshamoon: I like the functions but they're still doing the same as the Bash one-liner I mentioned in the other comment.

src/connect.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+echo -e 'connect 5C:BA:37:26:6D:9A\nquit' | bluetoothctl
```
johnshamoon: the -e doesn't work on the actual vehicle.

johnshamoon: bluetoothctl also requires superuser. It's probably not good practice to require the user to remember the sudo every time, especially when the error you get for not having sudo is  "No default controller available." That's confusing and I wouldn't think "oh, I just forgot sudo." Maybe make it "sudo bluetoothctl"

src/connect.sh
```diff
@@ -0,0 +1,5 @@
...
+
+
```
johnshamoon: please remove all these blank lines.

src/disconnect.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+echo -e 'disconnect 5C:BA:37:26:6D:9A\nquit' | bluetoothctl
+
```
johnshamoon: Same issues here as the previous script. It would be nice if these were both in the same scripts and we could just decide what function to use by passing arguments.

src/connect.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+echo 'connect 5C:BA:37:26:6D:9A\nquit' | bluetoothctl
```
johnshamoon: This still needs the sudo. bluetoothctl will not be able to register the agent without superuser rights.

src/controllerbluetooth.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+option=$1
+
+if [option = connect]
```
johnshamoon: Was this tested? In Bash, [ is a command so it needs spaces around it. You're also doing assignment instead of testing for equality. Just because Bash lets you do `option = connect`, doesn't mean you shouldn't specify that connect should be "connect"

src/controllerbluetooth.sh
```diff
@@ -0,0 +1,10 @@
...
+    echo 'connect 5C:BA:37:26:6D:9A\nquit' | sudo bluetoothctl
+then
```
johnshamoon: Why are you doing if then instead of if else? This is always executing both connect and disconnect.

src/controllerbluetooth.sh
```diff
@@ -0,0 +1,10 @@
...
+if[option = disconnect]
```
johnshamoon: `if[option`

Please test your code before pushing it. This is such an easy syntax error to catch that the GitHub syntax highlighter caught it.

src/controllerbluetooth.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# This bash file connects and disconnects the Xbox One S controller to the
```
johnshamoon: Change this to something like:

```
# This script connects and disconnects a Bluetooth device.
# The script should be called with the following format:
#      ./bt_toggle.sh $BTCOMAND $BTADDR
```

src/controllerbluetooth.sh
```diff
@@ -0,0 +1,16 @@
...
+# Raspberry Pis depending on the argument passed using the controller's.
+# MAC address. If the argument passed is "connect" then the controller is
+# connected, if "disconnect" is passed then the controller is disconnected
+
+option=$1
+
+if [[ $option == "connect" ]]; then
+    echo 'connect 5C:BA:37:26:6D:9A' | sudo bluetoothctl
```
johnshamoon: Instead of hard coding the BTADDR, can we replace it with a command-line argument? That way the script can be called by just saying something like:

./script.sh connect 5C:BA:37:26:6D:9A

src/controllerbluetooth.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# This bash file connects and disconnects the Xbox One S controller to the
+# Raspberry Pis depending on the argument passed using the controller's.
+# MAC address. If the argument passed is "connect" then the controller is
+# connected, if "disconnect" is passed then the controller is disconnected
+
+option=$1
```
johnshamoon: Can we also change this to something more descriptive? Especially if we add another parameter for the BTADDR. Something like
`command=$1`

johnshamoon: And add another variable for the bluetooth address like
`address=$2`

PR #30, opened 2019-03-01
Update README.md

changnor:
Corrected sudo swapon /s to sudo swapon /swapfile as /s is not the name of the swapfile


johnshamoon:
View PR #42 instead.

PR #31, opened 2019-03-02
script to determine leader or follower functionality

StevenDropiewski:
.


johnshamoon:
[Work on role switching script](https://trello.com/c/uvz9QhEm/86-work-on-role-switching-script)

src/startup.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Startup:
+
+    def __init__(self):
+        self.tag = TagRecognition()
+        
+    def choose_role(self):
+        obj_data = self.tag.detect()
+        if not obj_data:
+            system('python leader.py')
```
wisambunni: This runs a shell and execute the string s a bash command. Doing it this way is highly inefficient, unreliable, and insecure. It should be able to initialize and execute the leader code in Python natively.

src/startup.py
```diff
@@ -0,0 +1,27 @@
...
+        else:
+            system('python follower.py')
```
wisambunni: Same as my comment above.

src/startup.py
```diff
@@ -14,9 +14,9 @@ def __init__(self):
     def choose_role(self):
         obj_data = self.tag.detect()
         if not obj_data:
-            system('python leader.py')
+            leader.main()
```
johnshamoon: This is better but I'd prefer actual logic to be here. I don't think this even works? Leader does not have a method called main. There is a method called main in leader.py, but it isn't part of the class.

src/startup.py
```diff
@@ -14,9 +14,9 @@ def __init__(self):
...
         else:
-            system('python follower.py')
+            follower.main()
```
johnshamoon: Same as above.

src/startup.py
```diff
@@ -0,0 +1,48 @@
+from leader import *
+from follower import *
+"""
```
johnshamoon: Separate imports and docstrings with two blank lines.

src/startup.py
```diff
@@ -0,0 +1,48 @@
...
+            leader.drive()
+        if code == 'left_trigger':
+            leader.set_speed(position)
+            leader.reverse()
+        if code == 'dpad_left_right':
+            leader.turn(position)
+
+
+def main():
+
```
johnshamoon: No blank line here. It doesn't improve readability.

johnshamoon:
I also think that main.py is a better script name for this.

src/startup.py
```diff
@@ -0,0 +1,48 @@
...
+    detected = False
+
+    if data:
+        detected = true
+
+    return detected
+
+
+def follower():
+    follower = Follower()
```
wisambunni: Variable name also contradicts with the function it's in which contradicts with the module it's using. Your code has follower set up as three different entities and it makes the code hard to follow or differentiate between what is what.

src/startup.py
```diff
@@ -0,0 +1,48 @@
...
+    while True:
+        follower.follow()
+
+
+def leader():
+    leader = Leader()
```
wisambunni: Variable name also contradicts with the function it's in which contradicts with the module it's using. Your code has leader set up as three different entities and it makes the code hard to follow or differentiate between what is what.

src/main.py
```diff
@@ -0,0 +1,42 @@
+from leader import Leader
+from follower import Follower
+from tagrec import TagRecognition
+from inputcontroller import InputController
+
+
+"""
+Main script to run follower and leader functionality
```
johnshamoon: Try to be more descriptive and concise in the first line. Also end in a period always.
Maybe something like "Script to decide leader and follower roles." Then explain more.

src/main.py
```diff
@@ -0,0 +1,42 @@
...
+
+Takes input from camera to determine if current vehicle
```
johnshamoon: takes input from **the** camera to determine if **the** current vehicle

src/main.py
```diff
@@ -0,0 +1,42 @@
...
+from follower import Follower
+from tagrec import TagRecognition
+from inputcontroller import InputController
+
+
+"""
+Main script to run follower and leader functionality
+
+Takes input from camera to determine if current vehicle 
+should lead or follow based on detecting an AR tag.
```
johnshamoon: **ARTag**

based on whether or not there is an ARTag in front of the vehicle.

src/main.py
```diff
@@ -0,0 +1,42 @@
...
+
+If an AR tag exists the follower functionality will run.
```
johnshamoon: Change to

"If an ARTag exists, the vehicle will become a follower vehicle."

src/main.py
```diff
@@ -0,0 +1,42 @@
...
+If an AR tag does not exist the leader functionality will run.
```
johnshamoon: Change to

"If an ARTag does not exist, the vehicle will become the leader vehicle."

src/main.py
```diff
@@ -0,0 +1,42 @@
...
+"""
+def main():
+
```
johnshamoon: Don't need a blank line here.

src/main.py
```diff
@@ -0,0 +1,42 @@
...
+    tag = TagRecognition()
+
+    if tag.detect() is not None:
+        follower = Follower()
+        while True:
+            follower.follow()
+
```
johnshamoon: Don't need a blank line here either.

src/main.py
```diff
@@ -0,0 +1,42 @@
+from leader import Leader
```
johnshamoon: These should be in alphabetical order.

src/main.py
```diff
@@ -0,0 +1,42 @@
...
+    tag = TagRecognition()
+
+    if tag.detect() is not None:
+        follower = Follower()
+        while True:
+            follower.follow()
+
+    else:
+        leader = Leader()
+        controller = InputController()
```
johnshamoon: I don't really like how this is essentially leader.main, but there isn't really much more it should be doing. I'm going to wrap this in a function similar to follower.follow(). Expect that pull request in a few minutes and adjust this accordingly please.

src/main.py
```diff
@@ -0,0 +1,29 @@
...
+
+    if tag.detect() is not None:
+        follower = Follower()
+        while True:
+            follower.follow()
+    else:
+        leader = Leader()
+        while True:
+            leader.lead()
+
```
johnshamoon: two lines here

src/main.py
```diff
@@ -0,0 +1,29 @@
...
+should lead or follow based on whether or not there exists an ARTag
+in front of the vehicle.
+
+If an ARTag exists, the vehicle will become a follower vehicle.
+If an ARTag does not exist, the vehicle will become the leader vehicle.
+"""
+def main():
+    tag = TagRecognition()
+
+    if tag.detect() is not None:
```
johnshamoon: I think a short comment above this would be helpful.

src/main.py
```diff
@@ -0,0 +1,36 @@
...
+in front of the vehicle.
+
+If an ARTag exists, the vehicle will become a follower vehicle.
+If an ARTag does not exist, the vehicle will become the leader vehicle.
+"""
+def main():
+    tag = TagRecognition()
+    
+
+    """
```
johnshamoon: These are docstrings. I meant comments like

\# If an ARTag is detected, there is a vehicle in front of this vehicle so it should follow them.
\# If an ARTag is not detected, there is no vehicle in front of this vehicle so it should be a leader.

Again, those are just quick thoughts. Change the language however you like.

StevenDropiewski: ah. gotcha

StevenDropiewski: pep8 online check doesn't like long comments, (above 79 characters) it is a bit difficult to get it to fully describe functionality within that limit for a single line.

src/main.py
```diff
@@ -0,0 +1,32 @@
...
+should lead or follow based on whether or not there exists an ARTag
+in front of the vehicle.
+
+If an ARTag exists, the vehicle will become a follower vehicle.
+If an ARTag does not exist, the vehicle will become the leader vehicle.
+"""
+def main():
+    tag = TagRecognition()
+
+    #If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
```
johnshamoon: Space between the \# and the first character please.

PR #32, opened 2019-03-03
Camera Panning

changnor:
work in progress, needs more testing


src/pancam.py
```diff
@@ -0,0 +1,22 @@
+from follower import Follower
```
johnshamoon: Why is this its own script? It will only ever be used by the Follower class. I'm not exactly sure how this would even be used in the Follower class as is.

src/pancam.py
```diff
@@ -0,0 +1,22 @@
...
+
+follower = Follower()
+
+def seek():
+    #tag_data = follower.detect()
+    turn = follower.turn()
```
wisambunni: What does this exactly hold? It looks like it's just giving a command as turn() doesn't return anything.

src/pancam.py
```diff
@@ -0,0 +1,22 @@
+from follower import Follower
+
+follower = Follower()
```
johnshamoon: The only time we should have global scope is for `if __name__ == '__main__'`. I'm still not sure what the point of this script is, but if this is how you meant to keep it, you should declare follower in main and pass it to seek. Again, I don't think that's how this script should be.

src/pancam.py
```diff
@@ -0,0 +1,22 @@
...
+
+def seek():
+    #tag_data = follower.detect()
```
johnshamoon: Remove all the commented out dead code. We're not shooting for line count here.

src/pancam.py
```diff
@@ -0,0 +1,22 @@
...
+    turn = follower.turn()
+
+    if follower.detect() == False:
```
wisambunni: It's more Pythonic (and goes well with the rest of the code) to do:
`if not follower.detect()`

src/pancam.py
```diff
@@ -0,0 +1,22 @@
...
+        print('Detect False')
+        follower._speed = follower.LEADER_MAX_SPEED
```
wisambunni: Variables with leading underscore should be treated as private. It's not okay to write to a private member variable.

src/pancam.py
johnshamoon: `follower.turn()` turns the wheels to `_turn_angle`. `_turn_angle` isn't set until `follower.detect()`.

src/pancam.py
```diff
@@ -0,0 +1,22 @@
...
+        turn = turn + 35
```
wisambunni: This variable doesn't seem to be used anywhere in the code. Turn what 35 and in which direction? Why is it 35 in the first place?

src/pancam.py
johnshamoon: This can be changed to just be
```
if follower.detect():
  # true code
else:
  # false code
```

src/pancam.py
```diff
@@ -0,0 +1,22 @@
+from follower import Follower
+
+follower = Follower()
+
+def seek():
+    #tag_data = follower.detect()
+    turn = follower.turn()
+
+    if follower.detect() == False:
+        print('Detect False')
```
johnshamoon: We shouldn't be printing anything ever.

src/pancam.py
johnshamoon: What is this code doing? If we don't see the tag, floor it? We should be stopping in this case. And why are you controlling the speed at all in this script? This script is for camera panning, but I don't see camera panning at all in this script.

src/pancam.py
johnshamoon: Why are you adding 35 to turn? What does this mean? You're also not doing anything with turn, so I don't understand what this script is doing?

src/pancam.py
```diff
@@ -0,0 +1,22 @@
...
+        follower._speed = follower.LEADER_MAX_SPEED
+        turn = turn + 35
+    else:
+        print("Detect True")
```
johnshamoon: Again, we shouldn't be printing and there shouldn't be any dead code commented out.

src/pancam.py
```diff
@@ -0,0 +1,22 @@
+from follower import Follower
+
+follower = Follower()
+
+def seek():
```
wisambunni: If I understood the code correctly, this checks if the car sees a tag or not. If it does not see a tag, then it floors the vehicle (by overriding a private variable). It does not pan the camera or do anything of that sort.

src/pancam.py
```diff
@@ -0,0 +1,22 @@
...
+    else:
+        print("Detect True")
+        #print(tag_data)
+
+def main():
+    while True:
+        seek()
+
+if __name__ == '__main__':
+    main()
```
johnshamoon: Try to avoid having this "No newline at end of file" warning.

src/pancam.py
wisambunni: Using global variables in the code should be avoided unless necessary.

src/pancam.py
wisambunni: Overall, This shouldn't be a separate file (or a separate class). Seeking should be a function inside of the follower class that accompanies the rest of the follower's functionalities.

johnshamoon:
This shouldn't be experimental. It should be a feature on the car.

changnor:
No review request at the moment, currently not ready for review.

changnor:
Need to work on turning wheels with camera. Refine camera panning angles.

src/follower.py
```diff
@@ -9,6 +9,7 @@
 from tagrec import TagRecognition
 import numpy as np
 
+import time
```
johnshamoon: Remove time if you're not using it (it doesn't look like you are)

src/follower.py
```diff
@@ -44,6 +45,10 @@ def __init__(self):
         self._turn_angle = 0
         self._decision = 0
 
+        self._turn_right = 0
```
johnshamoon: Delete these three variables. You're not using them.

src/follower.py
```diff
@@ -98,17 +100,26 @@ def convert_camera_angle(self):
         else:
             self._turn_angle = 90
 
-
     """
     Resets the camera to the default position.
 
     The default position is tilted to 120 degrees and panned to 90 degrees.
     """
+
```
johnshamoon: Can you get rid of this blank line between doc strings and function names? Here and everywhere else. I don't think it's in PEP8 and even if it is, it's not consistent with the rest of our files.

src/follower.py
```diff
@@ -98,17 +100,26 @@ def convert_camera_angle(self):
...
+    
     def reset_camera(self):
         self._camera.turn_down(120)
         self._camera.pan_servo.write(90)
 
+    """
+    Definitions for turning the camera left and right
+    """
 
+    def _turn_camera_left(self, angle):
```
johnshamoon: Can we also get rid of the _ in the function name here? This one is (I think) in PEP8, but for consistency (for now), let's get rid of it. We'll go back later and privatize everything.

src/follower.py
```diff
@@ -98,17 +100,26 @@ def convert_camera_angle(self):
...
+        self._camera.turn_left(angle)
```
johnshamoon: camera.turn_left and camera.turn_right do not take angles. They take steps. Please check the camera script for more details on this.

src/follower.py
```diff
@@ -117,32 +128,47 @@ def reset_camera(self):
...
     def detect(self):
-        obj_data = self._tag.detect()
-        detected = False
-
-        if obj_data:
-            detected = True
-            self._distance = obj_data['z']
-            self._turn_angle = np.degrees(obj_data['direction'])
-            self._decision = obj_data['decision']
+        tag_data = self._tag.detect()
```
johnshamoon: Why'd you rewrite all of this? I have a few comments about it:

1. Don't change the variable name here for no reason. It's detecting an object. That object just so happens to be a tag. Additionally, this was reviewed and merged with no complaints. Please revert this back to obj_data.
2. You changed the entire if statement below, but only the parameter of the if statement needed to be changed (if the name change was going through). This blows up the line count for no reason.
3. There's a reason I didn't return tag_data in the if statement. Now the function returns two different data types. That's confusing and terrible for a public facing function. The way I had it, it returned a boolean 100% of the time and only had one exit point. This is more maintainable and testable than returning the tag data. There's also no benefit to returning the tag_data here. We don't do anything with it. There's a reason the variables are being set in this function.
4. IF this was to be changed, it should be in its own pull request. It has nothing to do with camera panning.

Please get rid of these changes.

src/follower.py
```diff
@@ -117,32 +128,47 @@ def reset_camera(self):
...
+            return tag_data
+        else:
+            return False
 
+    """
+    Keeps the AR tag in view of the camera to avoid losing tag.
+    """
 
+    def track_tag(self):
+        tag_data = self.detect()
```
johnshamoon: Why are you detecting here? We already have all the data. This is going to add a lot more processing that we don't need.

src/follower.py
```diff
@@ -117,32 +128,47 @@ def reset_camera(self):
...
+        pan_angle = 5
```
johnshamoon: Again, we're not turning in angles. Check camera.py to see what it's doing.

src/follower.py
```diff
@@ -117,32 +128,47 @@ def reset_camera(self):
...
+        
+        if tag_data == False:
+            self.stop()
```
johnshamoon: This was already happening. This script isn't about finding the tag after we've lost it. It's about never losing the tag in the first place. Get rid of this part.

src/follower.py
```diff
@@ -117,32 +128,47 @@ def reset_camera(self):
...
+        elif tag_data['x'] < 0:
+            self._turn_camera_left(pan_angle)
+            self.turn()
```
johnshamoon: Shouldn't be calling turn here. This function should be called in turn()

src/follower.py
```diff
@@ -117,32 +128,47 @@ def reset_camera(self):
...
     """
+
     def follow(self):
         if self.detect():
             self.drive()
             self.turn()
         else:
-            self.stop()
-
+            self.track_tag()
```
johnshamoon: Change this back to self.stop() and move your code into turn().

src/follower.py
```diff
@@ -117,32 +128,47 @@ def reset_camera(self):
...
+            self._decision = tag_data['decision']
+            return tag_data
+        else:
+            return False
 
+    """
+    Keeps the AR tag in view of the camera to avoid losing tag.
+    """
 
+    def track_tag(self):
```
johnshamoon: Can we change this function name to something more accurate? You're not tracking the tag in this function. You're following the tag by panning. Maybe something like pan_camera

src/follower.py
changnor: removed

src/follower.py
changnor: deleted

src/follower.py
changnor: got rid of it

src/follower.py
changnor: got rid of _ in function name, will need to go through later and privatize for consistency.

src/follower.py
changnor: corrected parameter names

src/follower.py
changnor: corrected parameter names

src/follower.py
changnor: got rid of it

src/follower.py
changnor: calling function in turn()

src/follower.py
changnor: reverted all changes

src/follower.py
changnor: changed name to pan_camera

src/follower.py
```diff
@@ -7,6 +7,7 @@
 import picar
 
 from tagrec import TagRecognition
+from math import pi
```
johnshamoon: Remove this for two reasons:
1. You're not using it anywhere.
2. numpy has pi in it (np.pi)

src/follower.py
```diff
@@ -37,17 +38,16 @@ def __init__(self):
         self._camera = Camera()
         self.reset_camera()
 
-        self._tag = TagRecognition()
+        self._tag = TagRecognition(marker_length=0.025)
```
johnshamoon: This is already in the repo. It's from my last commit. Remove this change and the repo again.

src/follower.py
```diff
@@ -37,17 +38,16 @@ def __init__(self):
...
         self._speed = 0
 
         self._distance = 0
         self._turn_angle = 0
         self._decision = 0
-
+        self._yaw = 0
```
johnshamoon: Same with this.

src/follower.py
```diff
@@ -73,35 +73,60 @@ def stop(self):
 
     """
     Turn the wheels towards the last recognized object.
+    Pans the camera towards the last recognized object.
```
johnshamoon: Remove this and add it to the line above it.

`"Turns the wheels and pans the camera ..."`

src/follower.py
```diff
@@ -73,35 +73,60 @@ def stop(self):
...
     """
     def turn(self):
+        self.pan_camera()
         self.convert_camera_angle()
         self._fw.turn(self._turn_angle)
 
 
     """
-    Converts the camera's angle scale to the same scale as the wheels.
+    Follows the tag by panning camera towards the same direction as the wheels.
```
johnshamoon: That's not what this is doing at all. It's not turning it towards the wheels. It's turning it left or right at 15 degrees (that's a different problem)>

src/follower.py
```diff
@@ -73,35 +73,60 @@ def stop(self):
...
+    
+    If the direction of the object is more/less than the distance then set the camera 3 steps
```
johnshamoon: Few problems:

1. Why does this have anything to do with distance? It exclusively deals with angles.
2. Don't say 3 steps, say how many degrees. Steps could be anything. This was really confusing when we first found out SunFounder was using steps, so imagine how confused people reading our code would be. Our code should be everything that we wanted their code to be.
3. Why does `pan_camera` compare the turn angle to the distance from the vehicle? How do you compare a distance in meters to an angle in degrees?
4. `self._distance` is the **z** distance (the distance TO the vehicle). It looks like you wanted the x distance.

src/follower.py
```diff
@@ -73,35 +73,60 @@ def stop(self):
...
+    def pan_camera(self):
+        if self._turn_angle < self._distance:
+            self.turn_camera_left(3)
+        elif self._turn_angle > self._distance:
+            self.turn_camera_right(3)
+        else:
+            self.reset_camera() 
+
+
+    """
```
johnshamoon: Add a comment for each function. If you're going to describe the parameter, explain what a step is. Again, this is just as bad as SunFounder doing this. Someone looking at this would have no idea what a step is. I would change this to degrees and do the conversion stuff yourself until we rewrite this. Remove the parameter description. We can add them when we change our comments to follow more of a documentation style.

src/follower.py
```diff
@@ -73,35 +73,60 @@ def stop(self):
...
+    def turn_camera_left(self, step):
+        self._camera.turn_left(step)
+    
+    def turn_camera_right(self, step):
+        self._camera.turn_right(step)
+
 
+    """
+    Converts the camera's angle scale to the same scale as the wheels.
```
johnshamoon: Remove this line. It's the same as what the next paragraph explains. It also has nothing to do with your pull request, so don't touch it.

src/follower.py
```diff
@@ -73,35 +73,60 @@ def stop(self):
...
     leftmost angle. Everything to the right of center is positive ranging from
     (90, 135] with 135 being the rightmost angle.
-
     The wheels turn on a range of [45, 135] with 45 being the rightmost, 135
     being the leftmost, and 90 being center.
     """
     def convert_camera_angle(self):
         if self._decision == -1:
-            self._turn_angle *= -1
+            self._turn_angle = np.abs(self._yaw - 90)
```
johnshamoon: Same as earlier, this is already merged and in the repository. Remove these changes and pull the changes from master.

src/follower.py
```diff
@@ -73,35 +73,60 @@ def stop(self):
...
         elif self._decision == 1:
-            self._turn_angle = 180 - self._turn_angle
+            self._turn_angle = 90 + self._yaw
```
johnshamoon: Same as above.

src/follower.py
```diff
@@ -125,13 +149,13 @@ def detect(self):
             self._distance = obj_data['z']
             self._turn_angle = np.degrees(obj_data['direction'])
             self._decision = obj_data['decision']
+            self._yaw = obj_data['yaw']
```
johnshamoon: Same as above.

src/follower.py
```diff
@@ -151,4 +175,4 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
+    main()
```
johnshamoon: Remove this.

src/follower.py
```diff
@@ -37,17 +38,16 @@ def __init__(self):
...
 
         self._distance = 0
         self._turn_angle = 0
         self._decision = 0
-
+        self._yaw = 0
 
     """
     Drives the vehicle forward and avoids collisions with recognized objects.
-
```
johnshamoon: Don't delete the blank line between the first line of the comment and the rest of the comment. It's there for a reason.  Get rid of this change here and everywhere else you did it.

src/follower.py
changnor: removed

src/follower.py
changnor: modified comment.

src/follower.py
changnor: re-pulled master

src/follower.py
changnor: removed

johnshamoon:
[Work on Pan and Tilt Following](https://trello.com/c/DLIv4FS6/83-work-on-pan-and-tilt-following)

src/follower.py
```diff
@@ -73,21 +72,53 @@ def stop(self):
...
+        self.pan_camera()
         self.convert_camera_angle()
         self._fw.turn(self._turn_angle)
 
 
+    """
+    Follows the tag by panning camera towards the same direction as the wheels.
+    """
+    def pan_camera(self):
+        if self._tag._tag_data['x'] < 0:
```
wisambunni: I don't think using the x value is a good way to do this. the x value is how far the tag is from the camera on the x axis. Using the direction property might make more sense for your case since we're looking to figure out how much should the camera pan by

src/follower.py
```diff
@@ -73,21 +72,53 @@ def stop(self):
...
+        else:
+            self.reset_camera() 
+
+
+    """
+    Turn the camera left and right.
+    Turn Left/Right functions from the Camera class only takes Steps.
+    """
+    def turn_camera_left(self):
+        self._camera.turn_left(self.angle_to_steps())
```
wisambunni: It would be better to pass the angle as a parameter (piece it with the direction from my comment above). This will let us control how sensitive we want the panning to be.

src/follower.py
```diff
@@ -73,21 +72,53 @@ def stop(self):
...
+    def turn_camera_right(self):
+        self._camera.turn_right(self.angle_to_steps())
+
+    
+    """
+    Convert angles to steps.
+    Takes the angle property and converts into steps where 1 step is 15 degrees.
+    """
+    def angle_to_steps(self):
+        steps = np.abs(self._turn_angle/self._camera.PAN_STEP)
```
wisambunni: Why do you need the absolute value if both numbers are always positive?

johnshamoon:
Please do not open a pull request unless the code is ready to be merged into master. Push often, but your commits shouldn't be "no review requested." They should be descriptive of what changed in this commit vs the previous commit stack. "No review requested" is so juxtaposed by the fact that it's in an actual pull request.

src/follower.py
```diff
@@ -37,18 +37,19 @@ def __init__(self):
         self._camera = Camera()
         self.reset_camera()
 
-        self._tag = TagRecognition(marker_length=0.025)
+        self._tag = TagRecognition(marker_length=0.60)
```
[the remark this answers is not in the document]
changnor: reverted

src/follower.py
```diff
@@ -87,7 +130,6 @@ def turn(self):
     to the left of center is negative ranging from [-45, -90) with -45 being the
     leftmost angle. Everything to the right of center is positive ranging from
     (90, 135] with 135 being the rightmost angle.
-
```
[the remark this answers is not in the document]
changnor: removed change

src/follower.py
```diff
@@ -111,7 +153,6 @@ def reset_camera(self):
 
     """
     Detects an ARTag and gets the object's data.
-
```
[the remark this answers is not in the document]
changnor: removed change

src/follower.py
```diff
@@ -132,14 +173,13 @@ def detect(self):
 
     """
     Follow an ARTag if one is found.
-
```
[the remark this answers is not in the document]
changnor: removed change

src/follower.py
changnor: used direction property

src/follower.py
changnor: Passed in angle as parameter with direction

src/follower.py
```diff
@@ -73,11 +74,53 @@ def stop(self):
...
+    """
+    Follows the tag by panning camera towards the same direction as the wheels.
+    """
+    def pan_camera(self):
+        if self._turn_angle < -1:
+            print('cameraL', self._turn_angle)
+            self.turn_camera_left(self._turn_angle)
+        elif self._turn_angle > 1:
+            self.turn_camera_right(self._turn_angle)
+            print('cameraR', self._turn_angle)
```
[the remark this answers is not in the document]
changnor: removed

src/follower.py
```diff
@@ -73,11 +74,53 @@ def stop(self):
...
     def turn(self):
-        self.convert_camera_angle()
-        self._fw.turn(self._turn_angle)
+        self.pan_camera()
+        self.prev_angles.append(abs(int(self._turn_angle)))
+        if len(self.prev_angles) > 10:
+          self.prev_angles.pop(0)
+          print(self.prev_angles[0], self.prev_angles[-1])
+        if self._turn_angle < -1 and (self.prev_angles[0] - 5 <= self.prev_angles[-1] <= self.prev_angles[0] + 5):
+            print('left')
```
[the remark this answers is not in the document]
changnor: removed

src/follower.py
```diff
@@ -73,11 +74,53 @@ def stop(self):
...
+            self._fw.turn(20)
+        elif self._turn_angle > 1 and (self.prev_angles[0] - 5 <= self.prev_angles[-1] <= self.prev_angles[0] + 5):
+            self._fw.turn(90)
+            print('right')
```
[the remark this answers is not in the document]
changnor: removed

src/follower.py
```diff
@@ -73,11 +74,53 @@ def stop(self):
...
+    Turn the wheels and pans the camera towards the last recognized object.
     """
     def turn(self):
-        self.convert_camera_angle()
-        self._fw.turn(self._turn_angle)
+        self.pan_camera()
+        self.prev_angles.append(abs(int(self._turn_angle)))
+        if len(self.prev_angles) > 10:
+          self.prev_angles.pop(0)
+          print(self.prev_angles[0], self.prev_angles[-1])
```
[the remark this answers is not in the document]
changnor: removed

src/follower.py
```diff
@@ -73,11 +74,53 @@ def stop(self):
...
+        elif self._turn_angle > 1 and (self.prev_angles[0] - 5 <= self.prev_angles[-1] <= self.prev_angles[0] + 5):
+            self._fw.turn(90)
+            print('right')
+
+    """
+    Follows the tag by panning camera towards the same direction as the wheels.
+    """
+    def pan_camera(self):
+        if self._turn_angle < -1:
+            print('cameraL', self._turn_angle)
```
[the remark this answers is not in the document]
changnor: removed

src/follower.py
```diff
@@ -73,11 +74,53 @@ def stop(self):
...
+            self._fw.turn(20)
+        elif self._turn_angle > 1 and (self.prev_angles[0] - 5 <= self.prev_angles[-1] <= self.prev_angles[0] + 5):
+            self._fw.turn(90)
+            print('right')
+
+    """
+    Follows the tag by panning camera towards the same direction as the wheels.
+    """
+    def pan_camera(self):
+        if self._turn_angle < -1:
```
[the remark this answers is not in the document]
changnor: reverted

src/follower.py
johnshamoon: This is only because you're calling pan_camera before convert_camera_angle

changnor: called pan_camera after convert_camera_angle. removed abs

src/follower.py
```diff
@@ -73,21 +71,57 @@ def stop(self):
...
 
     """
-    Converts the camera's angle scale to the same scale as the wheels.
+    Follows the tag by panning camera towards the same direction as the wheels.
+    """
+    def pan_camera(self):
+        increase_camera_offset = 0.5
+        if self._turn_angle < 90:
+            self.turn_camera_left(self._turn_angle)
+            self.camera_angle_offset = (90 - self._turn_angle)*increase_camera_offset
```
zeinyoussef: Why are you dividing the offset by 2? Shouldn't the offset be equal to the angle we pan the camera not just half of it? Also do the angles increase or decrease going toward the right from the car's point of view? In this case, if your turn angle is 75, then your offset would be 15. Is that correct or should it be -15?

src/follower.py
```diff
@@ -73,21 +77,63 @@ def stop(self):
...
     """
-    Turn the wheels towards the last recognized object.
+    Turn the wheels and pans the camera towards the last recognized object.
     """
     def turn(self):
         self.convert_camera_angle()
-        self._fw.turn(self._turn_angle)
+        self.pan_camera()
+
+        if self._turn_angle > 90 and self._turn_angle < self.wheel_max:
```
johnshamoon: python lets you do this way cleaner

90 < turn angle < wheel max

src/follower.py
```diff
@@ -45,10 +45,14 @@ def __init__(self):
         self._decision = 0
         self._yaw = 0
 
+        self.camera_angle_offset = 0
+
+        self.wheel_min = 45
```
johnshamoon: These should be constants.

src/follower.py
```diff
@@ -73,21 +77,63 @@ def stop(self):
...
+        
+    
+
+    """
+    Follows the tag by panning camera towards the same direction as the wheels.
+    """
+    def pan_camera(self):
+        if self._turn_angle < 90:
+            self.turn_camera_left(self._turn_angle)
+            self.camera_angle_offset = (90 - self._turn_angle)
```
johnshamoon: There's no scale conversion here. They're not on the same scale so you can't just subtract and call it good.

src/follower.py
```diff
@@ -73,21 +77,63 @@ def stop(self):
...
+        elif self._turn_angle > 90:
```
johnshamoon: What happens at 90?

src/follower.py
```diff
@@ -73,21 +77,63 @@ def stop(self):
...
+        self._camera.turn_right(self.angle_to_step())
 
+    
+    """
+    Definition for converting angles to steps.
+    Takes the angle property and converts into steps where 1 step is 5 degrees.
+    """
+    def angle_to_step(self):
+        steps = self._turn_angle/self._camera.PAN_STEP
+        return steps
```
johnshamoon: just return turn_angle/pan_step

src/follower.py
changnor: made into constants

src/follower.py
changnor: set to just return turn_angle/pan_step

src/follower.py
changnor: created scale conversion function

src/follower.py
johnshamoon: no you didn't

src/follower.py
johnshamoon: Where?

changnor: its in the newest commit

src/follower.py
changnor: I adjusted it in the newest commit

src/follower.py
```diff
@@ -87,21 +94,56 @@ def stop(self):
...
+        if self.STRAIGHT_ANGLE - self.DEADZONE_WHEEL < self._turn_angle:
+            self._fw.turn(self._turn_angle + self.camera_angle_offset) 
+        elif self.STRAIGHT_ANGLE + self.DEADZONE_WHEEL > self._turn_angle:
+            self._fw.turn(self._turn_angle - self.camera_angle_offset)
 
+    
     """
-    Converts the camera's angle scale to the same scale as the wheels.
+    Follows the tag by panning camera towards the same direction as the wheels.
+    """
```
wisambunni: Follow the new Sphinx style documentation for comments.

src/follower.py
```diff
@@ -87,21 +94,56 @@ def stop(self):
...
+    def pan_camera(self):
+
```
wisambunni: No need for a blank line here.

src/follower.py
```diff
@@ -87,21 +94,56 @@ def stop(self):
...
     """
-    Turn the wheels towards the last recognized object.
+    Turn the wheels and pans the camera towards the last recognized object.
     """
     def turn(self):
         self.convert_camera_angle()
-        self._fw.turn(self._turn_angle)
+        self.pan_camera()
 
+        if self.STRAIGHT_ANGLE - self.DEADZONE_WHEEL < self._turn_angle:
```
wisambunni: Shouldn't this be straight angle + deadzone?

wisambunni: Add comments for this if block that says which condition makes the vehicle turn left or right so it's easier to follow.

src/follower.py
```diff
@@ -87,21 +94,56 @@ def stop(self):
...
+            self._fw.turn(self._turn_angle + self.camera_angle_offset) 
+        elif self.STRAIGHT_ANGLE + self.DEADZONE_WHEEL > self._turn_angle:
```
wisambunni: Same thing here, shouldn't this be angle - deadzone?

changnor: changed

src/follower.py
changnor: Added comments

changnor: Yes, changed

src/follower.py
changnor: removed blank line

src/follower.py
changnor: Resolved merge conflicts and pulled latest version with Sphinx style documentation. Copied documentation style for panning.

src/follower.py
```diff
@@ -87,21 +94,56 @@ def stop(self):
...
+    
     """
-    Converts the camera's angle scale to the same scale as the wheels.
+    Follows the tag by panning camera towards the same direction as the wheels.
+    """
+    def pan_camera(self):
+
+        if self._turn_angle < self.STRAIGHT_ANGLE - self.DEADZONE_CAMERA:
+            self.turn_camera_left(self._turn_angle)
+            self.camera_angle_offset = self._camera.current_pan - 90
```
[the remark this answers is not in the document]
changnor: changed to np.abs(90 - self._camera.current_pan)

src/follower.py
```diff
@@ -66,11 +66,18 @@ def __init__(self):
         self._tag_lost_time = 0
         self._speed_cycle_time = time()
 
+        self.camera_angle_offset = 0
+
+        self.STRAIGHT_ANGLE = 90
+        self.WHEEL_MAX = 135
```
wisambunni: This variable needs to be more descriptive. Wheel max what? The radius of the wheel itself? The angle it can turn? The speed it can spin at?

src/follower.py
```diff
@@ -66,11 +66,18 @@ def __init__(self):
...
+        self.WHEEL_MIN = 45
```
wisambunni: See my comment above.

src/follower.py
```diff
@@ -66,11 +66,18 @@ def __init__(self):
...
+        self.DEADZONE_CAMERA = 7
```
wisambunni: See my comment above.

src/follower.py
```diff
@@ -66,11 +66,18 @@ def __init__(self):
...
+        self.DEADZONE_WHEELS = 10
```
wisambunni: See my comment above.

src/follower.py
```diff
@@ -106,11 +113,71 @@ def stop(self):
...
+            self.turn_camera_left(self._turn_angle)
+            self.camera_angle_offset = np.abs(90 - self._camera.current_pan)
+
+        elif self._turn_angle > self.STRAIGHT_ANGLE + self.DEADZONE_CAMERA:
+            # Right condition: If the turn angle is greater than 90 + deadzone, turn the camera right. 
+            # The offset is the current angle - 90.
+            self.turn_camera_right(self._turn_angle)
+            self.camera_angle_offset = np.abs(self._camera.current_pan - 90)
+
+        if self.camera_angle_offset == self.STRAIGHT_ANGLE or self.camera_angle_offset == self.WHEEL_MIN:
```
wisambunni: Why would the wheel be straight? As far as I understand WHEEL_MIN is set to 45 which means the maximum left turn angle for the vehicle.

src/follower.py
```diff
@@ -106,11 +113,71 @@ def stop(self):
...
+            # If the offset of the right or left offsets are equal to 90 or 45 respectively, turn the camera straight since the wheels will be straight.
+            self._camera.current_pan = self.STRAIGHT_ANGLE
+            self._camera.pan_servo.write(self._camera.current_pan)
+            self._fw.turn(self._camera.current_pan)
```
wisambunni: Why aren't you updating the camera offset at the end? It will still use the old camera offset for at least one cycle.

src/follower.py
```diff
@@ -106,11 +113,71 @@ def stop(self):
...
+        """
+        self._camera.turn_right(self.angle_to_step())
 
+    
+    def angle_to_step(self):
+        """
+        Definition for converting angles to steps. 
+        Takes the angle property and converts into steps where 1 step is 5 degrees.
+        """
+        return (self._turn_angle/self._camera.PAN_STEP) + 4
```
wisambunni: What is the purpose of adding a 4 here? Doesn't seem right.

src/follower.py
```diff
@@ -66,11 +66,18 @@ def __init__(self):
         self._tag_lost_time = 0
         self._speed_cycle_time = time()
 
+        self.camera_angle_offset = 0
```
wisambunni: Why isn't this a private member? No outside class should have access to the panning angle.

src/follower.py
changnor: I added 4 (magic number) to get the camera to turn a bit more than what it was already doing.

src/follower.py
changnor: I believe there was a request above to remove the privatization then go back and put them in, but I can do that now.

PR #36, opened 2019-03-07
Added experimental function to check if specific bluetooth address is…

StevenDropiewski:
… connected. Also allowing signals to go through bash to python if needed. To test, modify BTADDR in bt_check.py to your device's address.


src/isconnected.py
```diff
@@ -0,0 +1,14 @@
+import subprocess
```
johnshamoon: Still not sure why this is its own script. Where does this belong?

src/isconnected.py
```diff
@@ -0,0 +1,14 @@
...
+
+"""
+The method is_connected takes the output of bt_check.sh and returns a boolean value.
+The method is_connected is called with one argument, the bluetooth address of the paired device.
+"""
+def is_connected(BluetoothAddress):
```
johnshamoon: Check PEP8 for parameter naming styles.

src/isconnected.py
```diff
@@ -0,0 +1,14 @@
...
+    BTADDR = BluetoothAddress
```
johnshamoon: This variable is useless. Just pass the other parameter to subprocess.

src/isconnected.py
```diff
@@ -0,0 +1,14 @@
...
+    # This calls the isconnected.sh and gets a return value based on 
+    # the exit value in isconnected.sh
+    if subprocess.call(['./isconnected.sh', BTADDR]) == 1:
```
johnshamoon: you can just do

`return subprocess.call(['./isconnected.sh', BTADDR]) == 1`

and it'll have the exact same effect.

src/isconnected.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
johnshamoon: This is a one liner (basically). I think this can just be in the subprocess call in Python. I don't think we need another file for it.

src/isconnected.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# This script checks if a Bluetooth device is connected at a specific address.
+# This script can be called in this format:
+# 	./is_connected.sh $BTADDR 
+
+address=$1
+
+if echo info $address | sudo bluetoothctl | grep -q 'Connected: yes'; then
```
johnshamoon: See if you can find a better way to do this. Piping the command to grep is slow and not a great solution. I'm sure there's a better way to do it without having to do string searching. Try using devices in bluetoothctl.

src/isconnected.py
```diff
@@ -0,0 +1,14 @@
+import subprocess
+
+"""
+The method is_connected takes the output of bt_check.sh and returns a boolean value.
```
johnshamoon: I know this comment is referring to is_connected, that's why it's a docstring right above it. Just say "this method"

src/isconnected.py
```diff
@@ -0,0 +1,14 @@
...
+The method is_connected is called with one argument, the bluetooth address of the paired device.
```
johnshamoon: `...with one argument: the Bluetooth ...`

(Bluetooth is a proper noun)

src/isconnected.py
johnshamoon: Actually it's probably fine to just `return subprocess.call(['./isconnected.sh', BTADDR])`

src/isconnected.sh
StevenDropiewski: devices in bluetoothctl only shows paired devices, making no distinction if they are available or even connected.

johnshamoon: There's other ways (outside of bluetoothctl) to find connected devices. I don't like this grepping method. It's not going to be horribly slow because there's only like 10 lines of output and regex is fast, but there's gotta be a better way.

johnshamoon:
[Work on a script to determine if a Bluetooth address is connected](https://trello.com/c/Wa8dPIYw/96-work-on-a-script-to-determine-if-a-bluetooth-address-is-connected)

PR #37, opened 2019-03-08
Roleswapping

zeinyoussef:


src/main.py
```diff
@@ -13,20 +14,48 @@
...
     tag = TagRecognition()
+    btAddress = '5C:BA:37:26:6D:9A'
 
     # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
     # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
     if tag.detect() is not None:
-        follower = Follower()
-        while True:
-            follower.follow()
+        isLeader = False
```
wisambunni: Variables should be snake case.

src/main.py
```diff
@@ -13,20 +14,48 @@
...
-        while True:
-            leader.lead()
+        isLeader = True
+        vehicle = Leader()
+        bt_toggle('connect', btAddress)
+
+# This loop makes vehicle move and if the vehicle is a leader and sees an ARTag it will become
+# a follower and disconnect from the bluetooth controller
+
+    startTime = time.time()
```
wisambunni: Variables should be snake case.

src/main.py
```diff
@@ -13,20 +14,48 @@
...
+                bt_toggle('disconnect', btAddress)
 
+        else:
+            vehicle.follow()
+            if tag.detect() is not None:
+                startTime = time.time()
+            if (time.time() - startTime) > 5:
+                bt_toggle('connect', btAddress)     
+                isLeader = True
+                vehicle = Leader()
```
wisambunni: You need two blank lines between functions.

src/main.py
```diff
@@ -13,20 +14,48 @@
 If an ARTag exists, the vehicle will become a follower vehicle.
 If an ARTag does not exist, the vehicle will become the leader vehicle.
 """
```
wisambunni: This docu string should be located at the very top (with your name on it) since it's explaining what the file itself is doing instead of the function below the comment.

src/leader.py
```diff
@@ -62,12 +62,17 @@ def turn_right(self):
...
 
-
-    def turn(self,position):
-        if position == -1:
-            self.turn_left()
-        elif position == 1:
-            self.turn_right()
+    def set_turn_angle(self,position):
+        self.fw.turn(self.STRAIGHT_ANGLE + 45  * position)
+
```
wisambunni: According to PEP8 style guide, you need to have two blank lines between functions.

src/leader.py
```diff
@@ -80,7 +85,9 @@ def lead(self):
             self.set_speed(position)
             self.reverse()
         if code == 'dpad_left_right':
-            self.turn(position)
+            self.turn(position,'dpad_left_right')
+        if code == 'left_stick_x':
+            self.turn(position,'left_stick_x')
```
wisambunni: Lines 87-90 can be compressed into a single line of code.

src/leader.py
```diff
@@ -62,12 +62,17 @@ def turn_right(self):
...
         self.fw.turn(self.STRAIGHT_ANGLE)
 
-
-    def turn(self,position):
-        if position == -1:
-            self.turn_left()
-        elif position == 1:
-            self.turn_right()
+    def set_turn_angle(self,position):
+        self.fw.turn(self.STRAIGHT_ANGLE + 45  * position)
```
wisambunni: Using magic numbers is not allowed. Why are you adding 45 to your turn?

src/inputcontroller.py
```diff
@@ -146,10 +146,9 @@ def get_input(self):
 
 
 def main():
-    controller = InputController(debug=True)
+    controller = InputController()
```
johnshamoon: keep debug mode on

src/leader.py
johnshamoon: also PEMDAS. Is this doing what you want? Maybe wrap it in parentheses to make it clearer.

src/leader.py
```diff
@@ -62,12 +62,17 @@ def turn_right(self):
...
+    
+    def turn(self,position,code):
+        if code == 'dpad_left_right':
+            if position == -1:
+                self.turn_left()
+            elif position == 1:
+                self.turn_right()
```
johnshamoon: You still need to check if position == 0 to straighten the wheels. As it is now, the wheels will never straighten. You only call this function if the code is dpad_left_right or left_stick_x, so it will never fall into this else block.

src/leader.py
```diff
@@ -63,13 +63,21 @@ def turn_straight(self):
...
 
-    def turn(self,position):
-        if position == -1:
-            self.turn_left()
-        elif position == 1:
-            self.turn_right()
-        else:
-            self.turn_straight()
+    def set_turn_angle(self,position):
+        self.fw.turn(89 + (45  * position))
```
johnshamoon: What is 89? What is 45? Wisam asked you to remove the magic numbers.

src/leader.py
```diff
@@ -63,13 +63,21 @@ def turn_straight(self):
...
 
 
-    def turn(self,position):
-        if position == -1:
-            self.turn_left()
-        elif position == 1:
-            self.turn_right()
-        else:
-            self.turn_straight()
+    def set_turn_angle(self,position):
```
johnshamoon: Can you add a space between the comma and the second parameter.

src/leader.py
```diff
@@ -63,13 +63,21 @@ def turn_straight(self):
...
+        self.fw.turn(89 + (45  * position))
+
+    
+    def turn(self,position,code):
```
johnshamoon: Same here

src/main.py
```diff
@@ -13,19 +10,57 @@
...
+
```
johnshamoon: Remove this blank line.

src/main.py
```diff
@@ -13,19 +10,57 @@
...
+
+
+from follower import Follower
+from leader import Leader
+from tagrec import TagRecognition
```
johnshamoon: I think there's supposed to be a blank line between our imports and Python imports.

src/main.py
```diff
@@ -13,19 +10,57 @@
...
+import subprocess
+import time
+
+
+# This function calls the ./bt_toggle script with the specified command and address
```
johnshamoon: Use a doc string comment to describe functions.

src/main.py
```diff
@@ -13,19 +10,57 @@
...
     # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
     # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
     if tag.detect() is not None:
-        follower = Follower()
-        while True:
-            follower.follow()
+        bt_toggle('disconnect', bt_address)
+        is_leader = False
+        vehicle = Follower()
+
```
johnshamoon: Remove this blank line.

src/main.py
```diff
@@ -13,19 +10,57 @@
...
+    
     else:
-        leader = Leader()
-        while True:
-            leader.lead()
+        bt_toggle('connect', bt_address)
+        is_leader = True
+        vehicle = Leader()
+
+# This loop makes vehicle move and if the vehicle is a leader and sees an ARTag it will become
```
johnshamoon: Tab these comments in and try to limit the line length to closer to 80 characters.

src/main.py
```diff
@@ -13,19 +10,57 @@
...
+
+    start_time = time.time()
+    while True:
+        if is_leader:
+            vehicle.lead()
+            if tag.detect() is not None:
+                bt_toggle('disconnect', bt_address)
+                is_leader = False
+                vehicle = Follower()
+
```
johnshamoon: Remove this blank line

src/main.py
```diff
@@ -1,9 +1,6 @@
-from follower import Follower
-from leader import Leader
-from tagrec import TagRecognition
-
-
 """
+Written By: Zein Youssef and Steven Dropiewski
```
johnshamoon: Remove this for now. I'll give you a better format for this.

src/main.py
```diff
@@ -13,19 +10,57 @@
...
-        leader = Leader()
-        while True:
-            leader.lead()
+        bt_toggle('connect', bt_address)
+        is_leader = True
+        vehicle = Leader()
+
+# This loop makes vehicle move and if the vehicle is a leader and sees an ARTag it will become
+# a follower and disconnect from the bluetooth controller
+
```
johnshamoon: Remove this blank line.

src/leader.py
zeinyoussef: 89 is the angle where the car faces forwards and the 45 degrees is the largest turn the car can make, so I multiplied that value by the position which is a decimal from -1 to 1 in order to get teh angle we turn at. Also what do you mean by magic number?

src/leader.py
```diff
@@ -63,13 +63,21 @@ def turn_straight(self):
...
 
-    def turn(self,position):
-        if position == -1:
-            self.turn_left()
-        elif position == 1:
-            self.turn_right()
-        else:
-            self.turn_straight()
+    def set_turn_angle(self, position):
+        self.fw.turn(self.STRAIGHT_ANGLE + (self.MAX_TURN_ANGLE  * position))
```
johnshamoon: This function needs a comment. I know what it's doing because I wrote the InputController, but no one would know just by looking at it.

johnshamoon: Actually, just scrap this function. You're not setting the turn angle. you're turning the car itself. Just use this line of code instead of calling the function. We're not in it for LOC here.

src/main.py
```diff
@@ -13,19 +8,55 @@
...
 
     # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
     # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
     if tag.detect() is not None:
-        follower = Follower()
-        while True:
-            follower.follow()
+        bt_toggle('disconnect', bt_address)
+        is_leader = False
+        vehicle = Follower()
```
johnshamoon: You have like 4 blank spaces after the (). Get rid of those.

src/main.py
```diff
@@ -13,19 +8,55 @@
...
+    subprocess.call(['./bt_toggle.sh', command, address])
+
+
 def main():
     tag = TagRecognition()
+    bt_address = '5C:BA:37:26:6D:9A'
 
     # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
     # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
     if tag.detect() is not None:
```
johnshamoon: Can we swap this to
```
if tag.detect():
  detect stuff
else:
  not detect stuff

File: src/main.py [hunk 24]
```diff
@@ -13,19 +8,55 @@
...
+        vehicle = Follower()    
     else:
-        leader = Leader()
-        while True:
-            leader.lead()
+        bt_toggle('connect', bt_address)
+        is_leader = True
+        vehicle = Leader()
+
+    # This loop makes vehicle move and if the vehicle is a
```
johnshamoon: This is like 50 characters. Bring each line up to 80ish.

src/main.py
```diff
@@ -13,19 +8,55 @@
...
+        vehicle = Leader()
+
+    # This loop makes vehicle move and if the vehicle is a 
+    # leader and sees an ARTag it will become a follower and
+    # disconnect from the bluetooth controller
+    start_time = time.time()
+    while True:
+        if is_leader:
+            vehicle.lead()
+            if tag.detect() is not None:
```
johnshamoon: `if tag.detect():`

src/main.py
```diff
@@ -13,19 +8,55 @@
...
-            leader.lead()
+        bt_toggle('connect', bt_address)
+        is_leader = True
+        vehicle = Leader()
+
+    # This loop makes vehicle move and if the vehicle is a 
+    # leader and sees an ARTag it will become a follower and
+    # disconnect from the bluetooth controller
+    start_time = time.time()
+    while True:
```
wisambunni: While this whole block does what we want, it does it completely backwards. Why is it setting the timer if it sees the tag? All of this should be taken care of when we don't see a tag. it should start the timer when it doesn't see the tag and then build the rest of the algorithm around that.

src/main.py
```diff
@@ -13,19 +8,55 @@
...
+        if is_leader:
+            vehicle.lead()
+            if tag.detect() is not None:
+                bt_toggle('disconnect', bt_address)
+                is_leader = False
+                vehicle = Follower()
+        else:
+            vehicle.follow()
+            if tag.detect() is not None:
```
wisambunni: `if tag.detect():`

src/main.py
```diff
@@ -13,19 +8,55 @@
...
+                start_time = time.time()
+            if (time.time() - start_time) > 5:
+                bt_toggle('connect', bt_address)
```
wisambunni: How are you checking that the vehicle did in fact connect to the bluetooth controller?

src/main.py
```diff
@@ -13,19 +8,55 @@
...
-        while True:
-            follower.follow()
+        bt_toggle('disconnect', bt_address)
+        is_leader = False
+        vehicle = Follower()    
     else:
-        leader = Leader()
-        while True:
-            leader.lead()
+        bt_toggle('connect', bt_address)
```
wisambunni: How are you checking that the vehicle did in fact connect to the bluetooth controller?

src/main.py
```diff
@@ -13,19 +8,55 @@
...
     tag = TagRecognition()
+    bt_address = '5C:BA:37:26:6D:9A'
 
     # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
     # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
     if tag.detect() is not None:
-        follower = Follower()
-        while True:
-            follower.follow()
+        bt_toggle('disconnect', bt_address)
```
wisambunni: How are you checking that the vehicle did in fact disconnect from the Bluetooth controller?

src/main.py
zeinyoussef: I am waiting on the function that Steven is writing in order to make the checks.

src/main.py
zeinyoussef: Again, I am waiting on Steven to finish the is_connected function

src/main.py
zeinyoussef: The problem is if I start the timer when I don't see a tag then I have to add another loop within this while loop since the timer would be inside of it, this way I felt like it was cleaner and still does exactly what we need. I did remove the start_time = time.time() line before the loop though as it was unnecessary.

wisambunni: Sure we can go with that, but I think we need to have timer above the while loop still because say the vehicle never sees the tag in front of it, it will throw an error saying "undefined variable start_time" because it never gets initialized in the tag.detect() if statement.

src/main.py
```diff
@@ -13,19 +8,55 @@
...
-        while True:
-            leader.lead()
+        bt_toggle('connect', bt_address)
+        is_leader = True
+        vehicle = Leader()
+
+    # This loop makes vehicle move and if the vehicle is a leader and sees an ARTag
+    # it will become a follower and disconnect from the bluetooth controller
+    while True:
+        if is_leader:
```
johnshamoon: Instead of using an extra bool variable, couldn't we just check the type of vehicle?

src/main.py
zeinyoussef: ok I can add that in there

src/main.py
zeinyoussef: Yes we could but is there a way to check the type of object a variable is? Would the function type() work?

src/main.py
johnshamoon: is_connected is merged.

src/main.py
johnshamoon: http://bfy.tw/MkkP

src/main.py
```diff
@@ -14,28 +27,54 @@ def is_controller_connected():
...
+    # to the bluetooth controller. If done successfully it will become a leader, otherwise
+    # it will continue as a follower and reset the 5 second timer before trying to
+    # connect to the bluetooth controller again.
+    start_time = time.time()
+    while True:
+        if isinstance(vehicle, Leader):
+            vehicle.lead()
+            if tag.detect():
+                bt_toggle('disconnect', bt_address)
+                if not is_controller_connected():
```
johnshamoon: I don't think Bluetooth is going to work this quickly

src/main.py
```diff
@@ -14,28 +27,54 @@ def is_controller_connected():
...
+                    vehicle = Follower()
+        else:
+            vehicle.follow()
+            if tag.detect():
+                start_time = time.time()
```
johnshamoon: Why do we start a timer if we see a tag? Shouldn't we be resetting a timer if we don't see a tag?

src/main.py
```diff
@@ -14,28 +27,54 @@ def is_controller_connected():
...
+            if (time.time() - start_time) > 5:
+                bt_toggle('connect', bt_address)     
+                if is_controller_connected():
```
johnshamoon: Again, Bluetooth probably won't be this quick, so I'm pretty sure 99% of the time, it'll go into this else condition. The else condition is going to reset the timer and exit the top level if, trying to look for a tag again. Then it'll just cycle

src/main.py
```diff
@@ -14,28 +27,54 @@ def is_controller_connected():
...
-        leader = Leader()
-        while True:
-            leader.lead()
+        bt_toggle('connect', bt_address)
+    if is_controller_connected():
+        vehicle = Leader()
+    else:
+        vehicle = Follower()
+
+    # This loop makes vehicle move. If the vehicle is a leader and sees an ARTag
```
johnshamoon: This is way too verbose. It takes longer to read this than it does to read the code. This should be like two lines.

src/main.py
```diff
@@ -14,28 +27,54 @@ def is_controller_connected():
...
     tag = TagRecognition()
+    bt_address = '5C:BA:37:26:6D:9A'
 
-    # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
-    # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
-    if tag.detect() is not None:
-        follower = Follower()
-        while True:
-            follower.follow()
+    # If an ARTag is detected, the vehicle will try to disconnect from a bluetooth controller and
```
johnshamoon: Too verbose, repetitive, and exceeds the line length. There's also typos.

src/main.py
```diff
@@ -14,28 +27,54 @@ def is_controller_connected():
...
-Takes input from the camera to determine if the current vehicle
-should lead or follow based on whether or not there exists an ARTag
-in front of the vehicle.
 
-If an ARTag exists, the vehicle will become a follower vehicle.
-If an ARTag does not exist, the vehicle will become the leader vehicle.
-"""
 def main():
     tag = TagRecognition()
+    bt_address = '5C:BA:37:26:6D:9A'
```
johnshamoon: should be constant

src/main.py
zeinyoussef: This changes the start time to the current time any time we see a tag, so if a tag is absent for over 5 seconds then the code for connecting to a controller will run.

src/main.py
zeinyoussef: Have you tested to make sure? Does the connection line not pause the rest of the system? Also I can't think of a fix for this without adding a sleep for 1 second to give the system a chance to connect/disconnect successfully

johnshamoon: I haven't tested it because I know it won't work that fast. The connection line will not pause the rest of the system. It'll send the command and continue execution. I'm not sure of a better way either. Let me think about it.

src/main.py
johnshamoon: This is backwards. We shouldn't change the start time when we see a tag. We should change the start time when we don't see a tag.

zeinyoussef: The problem is that time.time() only goes upwards so this is the best way to do it without working with threads to make a timer. The point of setting start_time to time.time() is to reset the 5 second counter every time a tag is in frame.

src/main.py
```diff
@@ -14,28 +27,56 @@ def is_controller_connected():
...
+                if not is_controller_connected():
+                    vehicle = Follower()
+        else:
+            vehicle.follow()
+            if tag.detect():
+                timer_set = False
+            elif tag.detect() is None and timer_set == False:
+                start_time = time.time()
+                timer_set = True
+            if (time.time() - start_time) > 5:
```
wisambunni: Why aren't you setting the timer_set to False again?

src/main.py
```diff
@@ -14,28 +27,56 @@ def is_controller_connected():
...
+                bt_toggle('disconnect', BT_ADDRESS)
+                time.sleep(1)
+                if not is_controller_connected():
+                    vehicle = Follower()
+        else:
+            vehicle.follow()
+            if tag.detect():
+                timer_set = False
+            elif tag.detect() is None and timer_set == False:
+                start_time = time.time()
```
wisambunni: start_time should be defined outside of the condition blocks. Your code will throw an error if it sees a tag and then tries to count using an (undefined) variable in line 73.

src/main.py
```diff
@@ -14,28 +27,58 @@ def is_controller_connected():
...
+            if tag.detect():
+                bt_toggle('disconnect', BT_ADDRESS)
+                time.sleep(1)
+                if not is_controller_connected():
+                    vehicle = Follower()
+        else:
+            vehicle.follow()
+            if tag.detect():
+                timer_set = False
+            elif not tag.detect() and timer_set == False:
```
johnshamoon: Be consistent here. You used `not tag.detect()` and `and timer_set` elsewhere, so use `not timer_set` here

src/main.py
```diff
@@ -14,28 +27,58 @@ def is_controller_connected():
...
     tag = TagRecognition()
+    BT_ADDRESS = '5C:BA:37:26:6D:9A'
 
-    # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
-    # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
-    if tag.detect() is not None:
-        follower = Follower()
-        while True:
-            follower.follow()
+    # If an ARTag is detected, the vehicle will disconnect from a bluetooth
```
johnshamoon: Bluetooth*

johnshamoon:
[Finish Role Swapping](https://trello.com/c/uQuOG6et/102-finish-role-swapping)

src/bluetoothctl.py
```diff
@@ -0,0 +1,204 @@
+# ReachView code is placed under the GPL license.
```
johnshamoon: Don't add this file yet (to git). I'm still figuring out how we're going to use this and I already have a branch that adds it (I just haven't opened a pull request yet). Go ahead and take a look at that branch. Ignore the leader.py changes, they shouldn't be there, but take a look at how I'm using it.

It also shouldn't be in this PR. Adding it to what we have right now should be a separate pull request because main.py currently does not work for the leader.

src/main.py
```diff
@@ -1,11 +1,25 @@
...
+
+If an ARTag exists, the vehicle will become a follower vehicle.
+If an ARTag does not exist, the vehicle will become the leader vehicle.
+"""
 
 from follower import Follower
 from leader import Leader
 from tagrec import TagRecognition
 
+from bluetoothctl import *
```
johnshamoon: this would be `from bluetoothctl import Bluetoothctl`

but again, it's in my other branch right now.

src/main.py
```diff
@@ -14,29 +28,55 @@ def is_controller_connected():
...
-    # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
+    # If an ARTag is detected, the vehicle will disconnect from a Bluetooth
+    # controller. If an ARTag is not detected, the vehicle will connect to
+    # a Bluetooth controller.
     if tag.detect():
-        tag = None
-        follower = Follower()
-        while True:
-            follower.follow()
+        while is_controller_connected():
```
johnshamoon: Comment applies to all the Bluetooth stuff:

not quite how I was thinking we'd use this. Take a look at my branch and see how I'm using it and validating that events actually happened.

johnshamoon: But again, I'm not sure it should be in this PR because we need to fix it before we put role swapping in.

zeinyoussef: Ok, I'll leave the connect and disconnect functions in there just as place holders until we decide how we are going to use the python wrapper

src/main.py
```diff
@@ -10,36 +10,74 @@
...
+
+    # This loop makes the vehicle move. If the vehicle sees an ARTag then
+    # it is a follower vehicle, otherwise it is a leader vehicle.
+    timer_set = False
+    start_time = time.time()
+    while True:
+        if isinstance(vehicle, Leader):
+            vehicle.lead()
+            if tag.detect():
+                while not is_controller_connected():
```
zeinyoussef: *while is_controller_connected()

src/main.py
```diff
@@ -10,36 +10,74 @@
...
+                    bl.disconnect('5C:BA:37:26:6D:9A')
+                vehicle = Follower()
+        else:
+            vehicle.follow()
+            if tag.detect():
```
johnshamoon: This is what breaks it. Every time we call tag.detect, we get a different result. That's not good. I think what we should do is return the data that tag.detect() in the follower gets, then use that everywhere.

src/main.py
```diff
@@ -10,36 +10,75 @@
...
+            vehicle.lead()
+            if tag_visible:
+                while is_controller_connected():
+                    bl.disconnect('5C:BA:37:26:6D:9A')
+                vehicle = Follower()
+        else:
+            vehicle.follow()
+            if tag_visible:
+                timer_set = False
+            elif not tag_visible and not timer_set:
```
johnshamoon: can't we just do `elif not timer_set` because we know for a fact that if we get here, there wasn't a tag

src/main.py
```diff
@@ -10,36 +10,75 @@
...
+        while not is_controller_connected():
+            bl.connect('5C:BA:37:26:6D:9A')
+        vehicle = Leader()
+
+    # This loop makes the vehicle move. If the vehicle sees an ARTag then
+    # it is a follower vehicle, otherwise it is a leader vehicle.
+    timer_set = False
+    start_time = time.time()
+    while True:
+        tag_visible = tag.detect()
```
johnshamoon: can we make this just `tag` so that it matches the rest of the code?

src/main.py
```diff
@@ -10,36 +10,75 @@
...
+            elif not tag_visible and not timer_set:
+                start_time = time.time()
+                timer_set = True
+            if (time.time() - start_time) > 5 and timer_set:
+                timer_set = False
+                bl.connect('5C:BA:37:26:6D:9A')
+                if is_controller_connected():
+                    vehicle = Leader()
+                else:
+                    start_time = time.time()
```
johnshamoon: Don't you need to set `timer_set` to `True` here?

src/main.py
```diff
@@ -10,36 +10,75 @@
...
+
+
 def main():
     tag = TagRecognition()
+    bl = Bluetoothctl()
+    BT_ADDRESS = '5C:BA:37:26:6D:9A'
 
-    # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
-    # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
+    # If an ARTag is detected, the vehicle will disconnect from a Bluetooth
```
johnshamoon: I'd leave this unchanged. I think saying that it becomes a leader or a follower (like it did) is more detailed. Otherwise, this can mean that if the leader sees an ARTag, it disconnects from Bluetooth but stays a leader. I think it was better before.

src/main.py
```diff
@@ -10,36 +10,75 @@
...
+
+
 def main():
     tag = TagRecognition()
+    bl = Bluetoothctl()
+    BT_ADDRESS = '5C:BA:37:26:6D:9A'
 
-    # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow.
-    # If an ARTag is not detected, no vehicle is present so this vehicle will be a leader.
+    # If an ARTag is detected, there is a vehicle ahead of this vehicle so it will follow
```
johnshamoon: Forgot the periods.

src/main.py
```diff
@@ -71,9 +70,39 @@ def main():
...
-            leader.lead()
+        vehicle = Leader()
+
+    # This loop makes the vehicle move. If the vehicle sees an ARTag then
+    # it is a follower vehicle, otherwise it is a leader vehicle.
+    timer_set = False
+    start_time = time.time()
+    while True:
+        tag_visible = tag.detect()
+        if isinstance(vehicle, Leader):
```
wisambunni: Turn this whole block into a bunch of chunks. It makes it really hard to read and follow the code without few blank lines to separate related code

src/main.py
```diff
@@ -71,9 +70,39 @@ def main():
...
 
-        leader = Leader()
-        while True:
-            leader.lead()
+        vehicle = Leader()
+
+    # This loop makes the vehicle move. If the vehicle sees an ARTag then
+    # it is a follower vehicle, otherwise it is a leader vehicle.
+    timer_set = False
+    start_time = time.time()
```
wisambunni: blank line here

experimental/role_swapping.py
```diff
@@ -0,0 +1,117 @@
...
+    :type bt_addr: string
+    """
+    bt.disconnect(bt_addr)
+    sleep(Follower.CYCLE_TIME)
+    if any(d['mac_address'] == bt_addr for d in bt.get_paired_devices()):
+        bt.remove(bt_addr)
+
+
+def main():
+    """
```
johnshamoon: Can you remove this docstring

experimental/role_swapping.py
```diff
@@ -0,0 +1,117 @@
+"""
+Script to determine leader and follower roles.
```
johnshamoon: Can you shorten this whole docstring to one line?
```
Experimental script for role swapping.

Author: Zein Youssef
```

PR #38, opened 2019-03-13
Customcamera

StevenDropiewski:
Camera api adding rudimentary ability to operate camera pan and tilt with absolute degree references


src/customcamera.py
```diff
@@ -0,0 +1,83 @@
...
+    '''
+    def tilt(self, degree):
+        self.tilt_servo.write(self.limit_tilt(degree))
+
+    '''
+    Writes tilt offset to a config file within current directory, lower offset directs the camera
+    more towards the ground while higher offset directs the camera more vertical. This file can
+    be changed incrementally with this function or manually in the files system.
+    '''
+    def calibrate_tilt(self, degree):
```
wisambunni: It would look cleaner if limit_tilt and limit pan were combined into one function, validate_degree(...) for an example.

StevenDropiewski: I agree, but I don't want the tilt to output 180 degrees, it may break the servo when it moves back that far.

src/camera.py
```diff
@@ -0,0 +1,85 @@
...
+'''
+Camera class.
+
+Uses SunFounder Servo class to control camera servos allowing for pan and tilt.
+Uses SunFounder filedb to configure offset for pan and tilt functions.
+'''
+class Camera:
+    PAN_CHANNEL = 1
+    TILT_CHANNEL = 2
+    DB = filedb.fileDB(getcwd() + "/../SunFounder_PiCar-V/remote_control/remote_control/driver/config")
```
johnshamoon: this should still be init, but as a constant.

johnshamoon: also private.

src/customcamera.py
johnshamoon: So don't do that, but still make it one function.

src/camera.py
```diff
@@ -0,0 +1,79 @@
...
+    '''
+    Returns the tilt and pan ouptut at maximum or minimum value if parameter entered exceeds 
+    listed maximum or minimum value for tilt and pan respectively. Second argument determines if
+    degree is pan or tilt, 1 for pan, 0 for tilt.
+
+    Inputs : Degree, pan or tilt.
+    Outputs : Validated degree.
+    '''
+    def validate_degree(self, degree, pan_or_tilt)
+        if degree > 180 and pan_or_tilt == 1:
```
johnshamoon: make this something like
```
if degree > max:
  degree = max
elif degree < 0:
  degree = 0
```

src/camera.py
```diff
@@ -0,0 +1,79 @@
...
+        self._DB.set('pan_offset', new_pan_cali)
+        self._pan_offset = new_pan_cali
+
+
+    '''
+    Controls the til movement of the camera. Usable inputs are [0,150]. 0 refers to 
+    the vehicles front while 150 refers to the camerafacing upwards.
+    '''
+    def tilt(self, degree):
+        self._tilt_servo.write(self.validate_degree(degree,0))
```
johnshamoon: Pass in the max value, not a flag.

src/camera.py
```diff
@@ -0,0 +1,76 @@
...
+'''
+class Camera:
+    PAN_CHANNEL = 1
+    TILT_CHANNEL = 2
+
+    def __init__(self, bus_number=1):
+        self._DB = filedb.fileDB(getcwd() + "/../SunFounder_PiCar-V/remote_control/remote_control/driver/config")
+        self._pan_offset = int(self._DB.get('pan_offset', default_value=0))
+        self._tilt_offset = int(self._DB.get('tilt_offset', default_value=0))
+        self._pan_servo = Servo.Servo(self.PAN_CHANNEL, bus_number=bus_number, offset=_self.pan_offset)
```
johnshamoon: Can you try to limit these lines to around 80 characters? Break them up at reasonable points like the commas or something.

src/camera.py
```diff
@@ -0,0 +1,76 @@
...
+        self._tilt_servo = Servo.Servo(self.TILT_CHANNEL, bus_number=bus_number, offset=_self.tilt_offset)
+
+
+    '''
+    Controls the horizontal movement of the camera.Usable inputs are [0,180]. 0 refers to the 
+    vehicles righthand side while 180 refers to the vehicles lefthand side.
```
johnshamoon:
```
Pans the camera.

The camera can pan in [0, 180] degrees. 0 refers to the left and 180 refers to the right.

File: src/camera.py [hunk 7]
```diff
@@ -0,0 +1,76 @@
...
+    or manually in the file system.
+    '''
+    def calibrate_pan(self, degree):
+        new_pan_cali = self._pan_offset + degree
+        self._DB.set('pan_offset', new_pan_cali)
+        self._pan_offset = new_pan_cali
+
+
+    '''
+    Controls the til movement of the camera. Usable inputs are [0,150]. 0 refers to
```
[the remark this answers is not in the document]
johnshamoon: The panning and tilting docstrings should basically be identical. They should have the same wording and the same point. Consistency is important.

src/camera.py
```diff
@@ -0,0 +1,76 @@
...
+        self._DB.set('pan_offset', new_pan_cali)
+        self._pan_offset = new_pan_cali
+
+
+    '''
+    Controls the til movement of the camera. Usable inputs are [0,150]. 0 refers to 
+    the vehicles front while 150 refers to the camerafacing upwards.
+    '''
+    def tilt(self, degree):
+        self._tilt_servo.write(self.validate_degree(degree,150))
```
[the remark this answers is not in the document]
johnshamoon: also add a space between the comma and the value.

src/camera.py
```diff
@@ -0,0 +1,76 @@
...
+
+    '''
+    Controls the horizontal movement of the camera.Usable inputs are [0,180]. 0 refers to the 
+    vehicles righthand side while 180 refers to the vehicles lefthand side. 
+    '''
+    def pan(self, degree):
+        self._pan_servo.write(self.validate_degree(degree,180))
+
+
+    '''
```
johnshamoon: This isn't true. It doesn't write to a config within the current directory. Also this isn't true. Adding an offset of 10 and an offset of 100 should both move the camera in the same direction because you're just adding the offset. This is what the SunFounder API does, which I told you not to copy. Our system should be absolute, not relative. Fix the comments to be true, fix the line lengths to fit 80 characters, and fix the function.

As it is, I can pass in 100000 and it'll take it.

src/camera.py
```diff
@@ -0,0 +1,76 @@
...
+    more towards the ground while higher offset directs the camera more vertical. This file can
+    be changed incrementally with this function or manually in the files system.
+    '''
+    def calibrate_tilt(self, degree):
+        new_tilt_cali = self._tilt_offset + degree
+        self._DB.set('tilt_offset', new_tilt_cali)
+        self._tilt_offset = new_tilt_cali
+
+
+    '''
```
johnshamoon: Doesn't return the tilt and pan **_ouptut_**. It validates a value is within [0, max]. Say that. Yes, it's used to validate the tilt and pan degrees are in the right ranges, but that's it's more generic than that.

Also, get rid of the inputs and outputs for now. Describe them in the comments. We'll have a better format for this later.

src/camera.py
```diff
@@ -0,0 +1,76 @@
...
+    Returns the tilt and pan ouptut at maximum or minimum value if parameter entered exceeds 
+    listed maximum or minimum value for tilt and pan. Second argument sets maximum value.
+
+    Inputs : Degree, Maximum.
+    Outputs : Validated degree.
+    '''
+    def validate_degree(self, degree, max)
+        if degree > max:
```
johnshamoon: you can do this in one line btw. Try that.

src/camera.py
```diff
@@ -0,0 +1,92 @@
...
+    def calibrate_tilt(self, degree):
+        valid_degree = validate_degree(degree, MIN_CALIBRATE, MAX_CALIBRATE)
+        self._DB.set('tilt_offset', valid_degree)
+        self._tilt_offset = valid_degree
+
+
+    '''
+    Validates degree. 
+
+    Validates a degree within [min_degree, max_degree]
```
johnshamoon: This comment is split in a weird place (like 50 characters). Bring it up to 80.

src/camera.py
```diff
@@ -0,0 +1,92 @@
...
+    Sets current tilt offset to new validated degree.
+    '''
+    def calibrate_tilt(self, degree):
+        valid_degree = validate_degree(degree, MIN_CALIBRATE, MAX_CALIBRATE)
+        self._DB.set('tilt_offset', valid_degree)
+        self._tilt_offset = valid_degree
+
+
+    '''
+    Validates degree.
```
johnshamoon: This comment is essentially useless. Make it a little more descriptive.

src/camera.py
```diff
@@ -0,0 +1,92 @@
...
+Uses servo class to control pan and tilt of the vehicles camera.
+Uses filedb to set and get offset values for both pan and tilt.
+'''
+class Camera:
+    PAN_CHANNEL = 1
+    TILT_CHANNEL = 2
+    MAX_PAN_DEGREE = 180
+    MAX_TILT_DEGREE = 150
+    MIN_DEGREE = 0
+    MAX_CALIBRATE = 50
```
johnshamoon: MAX_CALIBRATE what? Degrees?

src/camera.py
```diff
@@ -0,0 +1,92 @@
...
+    '''
+    def pan(self, degree):
+        self._pan_servo.write(self.validate_degree(degree, MIN_DEGREE, \
+        MAX_PAN_DEGREE))
+
+
+    '''
+    Calibrates camera pan.
+
+    Calibrates pan in [-50, 50].
```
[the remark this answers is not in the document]
johnshamoon: This wasn't me telling you to change the comment. I'm asking what it does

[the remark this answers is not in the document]
StevenDropiewski: the SunFounder calibrate and the way I wrote it here adds the offset value to the value generated by _angle_to_analog() within their write function. I'm going to rewrite it so it passes nothing to the offset but instead calibrates by adding the offset to the angle variable being passed to the servo write, thereby always keeping it between the required 0 and 180 and not arbitrarily adding values after converting for the pwm.

src/camera.py
```diff
@@ -0,0 +1,93 @@
...
+        self._tilt_servo = Servo.Servo(self.TILT_CHANNEL, 
+                                       bus_number=bus_number, 
+                                       offset= OFFSET_TO_SERVO)
+
+
+    '''
+    Pans the camera.
+
+    Controls the horizontal movement of the camera in [0, 180] degrees. 
+    0 refers to the vehicle's right while 180 refers to the vehicle's left.
```
StevenDropiewski: Should I add that it uses offset ?

src/camera.py
```diff
@@ -0,0 +1,93 @@
...
+        self._DB = filedb.fileDB(getcwd() + 
+                                 "/../SunFounder_PiCar-V/remote_control/remote_control/driver/config")
+        self._pan_offset = int(self._DB.get('pan_offset', default_value=0))
+        self._tilt_offset = int(self._DB.get('tilt_offset', default_value=0))
+        self._pan_servo = Servo.Servo(self.PAN_CHANNEL,  
+                                      bus_number=bus_number, 
+                                      offset= OFFSET_TO_SERVO)
+        self._tilt_servo = Servo.Servo(self.TILT_CHANNEL, 
+                                       bus_number=bus_number, 
+                                       offset= OFFSET_TO_SERVO)
```
StevenDropiewski: is the space here a problem?

src/camera.py
```diff
@@ -0,0 +1,105 @@
...
+'''
+class Camera:
+    PAN_CHANNEL = 1
+    TILT_CHANNEL = 2
+    MAX_PAN_DEGREE = 180
+    MAX_TILT_DEGREE = 150
+    MIN_DEGREE = 0
+    OFFSET_TO_SERVO = 0
+
+    def __init__(self, bus_number=1):
```
StevenDropiewski: will add bus_number to constants

src/camera.py
```diff
@@ -0,0 +1,105 @@
...
+
+    '''
+    Calibrates camera pan.
+
+    Controls offset in degrees for pan. Writes new offset value to
+    'SunFounder_PiCar-V/remote_control/remote_control/driver/config' 
+    to 'pan_offset'. 
+    '''
+    def calibrate_pan(self, degree):
+        self._DB.set('pan_offset', degree)
```
johnshamoon: Again, there's no validation here. I know you're validating later, but why put the system in an inconsistent state by writing wrong information to the config file.

Additionally, do you know how this config file is working? I understand that it _has_ those offsets, but _why_ does it have those offsets?

Lastly, if you're changing _DB, it's not a constant.

src/camera.py
```diff
@@ -0,0 +1,105 @@
...
+    Pans the camera.
+
+    Controls the horizontal movement of the camera in [0, 180] degrees. 
+    0 refers to the vehicle's right while 180 refers to the vehicle's left. 
+    '''
+    def pan(self, degree):
+        self._current_pan_angle = self.validate_degree(degree,
+                                                       self.MIN_DEGREE,
+                                                       self.MAX_PAN_DEGREE)
+        self._pan_servo.write(self.validate_degree(degree + self._pan_offset,
```
johnshamoon: Makes more sense to use _current_pan_angle instead of degree in this call.

src/camera.py
```diff
@@ -0,0 +1,105 @@
...
+    OFFSET_TO_SERVO = 0
+
+    def __init__(self, bus_number=1):
+        self._DB = filedb.fileDB(getcwd() + 
+                                 "/../SunFounder_PiCar-V/remote_control/remote_control/driver/config")
+        self._pan_offset = int(self._DB.get('pan_offset', default_value=0))
+        self._tilt_offset = int(self._DB.get('tilt_offset', default_value=0))
+        self._pan_servo = Servo.Servo(self.PAN_CHANNEL,  
+                                      bus_number=bus_number, 
+                                      offset=self.OFFSET_TO_SERVO)
```
johnshamoon: Why is this `OFFSET_TO_SERVO` instead of the value that you just grabbed?

src/camera.py
johnshamoon: No, because it changes depending on who calls it.

if I create an object like this: `camera = Camera(bus_number=10)`, I changed the value of bus number. Compare that functionality to that of the rest of the constants.

src/camera.py
johnshamoon: Do we also want that behavior though? I don't know if we should be messing with the PWM stuff if they need it.

src/camera.py
johnshamoon: Why not just validate to ensure that we can never write any value that could break the servo? Validate at every level that takes input from somewhere else that hasn't already been validated.

src/camera.py
```diff
@@ -0,0 +1,111 @@
...
+Calibration adds values to the pan() or tilt().
+'''
+class Camera:
+    PAN_CHANNEL = 1
+    TILT_CHANNEL = 2
+    MAX_PAN_DEGREE = 180
+    MAX_TILT_DEGREE = 150
+    MIN_DEGREE = 0
+    MAX_CALI_DEGREE = 180
+    MIN_CALI_DEGREE = -180
```
johnshamoon: what

StevenDropiewski: If you want to refer to 0 as 180 and work in 180 to 360, why not give that possibility? The way the validation works it won't break the servo, and otherwise I don't know what ranged to put for the calibration, it would all be arbitrary.

johnshamoon: I don't understand why we'd be working in 180 to 360 though?

StevenDropiewski: Me neither. But I had the option to calibrate between the range -50, 50 which seemed to be a decent middle ground. Would that be a better alternative?

johnshamoon: I mean, would it be so bad if we just didn't do any calibration? Clearly no one complained about their calibration functions not working.

StevenDropiewski: because it was using the values in the offset config file. You showed it a while ago, it was actually grabbing the offset values, just not rewriting them if you wanted to change them.

johnshamoon: Right, but those values are the default values for a reason. What do we gain by changing them?

StevenDropiewski: I believe having some flexibility and extra functionality on the camera would not be a hindrance. But as team lead and technical lead it is your decision.

src/camera.py
```diff
@@ -0,0 +1,88 @@
...
+"/../SunFounder_PiCar-V/remote_control/remote_control/driver/config"under
+pan_offset and tilt_offset. 
+
+'''
+class Camera:
+    PAN_CHANNEL = 1
+    TILT_CHANNEL = 2
+    MAX_PAN_DEGREE = 180
+    MAX_TILT_DEGREE = 150
+    MIN_DEGREE = 0
```
zeinyoussef: I thought the camera's max and min angles for panning were 45 and 135 degrees?

johnshamoon: That's for turning

src/camera.py
```diff
@@ -0,0 +1,212 @@
...
+        :param max_degree: the upper bound of the validation.
+        :type max_degree: float
+
+        :return: The value of degree bounded between the min and max parameters.
+        :rtype: float
+        """
+        return max(min(max_degree, degree), min_degree)
+
+
+    def test_pan(self):
```
johnshamoon: I don't think this needs to be a function of its own. Just throw the body in main. You can keep the docstring though, just put it in main too.

src/camera.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        Tests the panning of the camera
```
johnshamoon: `Demonstrates the camera's panning functionality.`

Also shouldn't this tilt too?

src/camera.py
```diff
@@ -0,0 +1,212 @@
...
+
+        sets the camera to 90 degrees then to right most value at 0 degrees.
```
johnshamoon: Not sure if we need to detail everything that it's doing. Just say that it turns from 0 to 180 and back or something.

src/camera.py
```diff
@@ -0,0 +1,212 @@
...
+        values showing validation function working as intended, ending at 180 
+        degrees. Then pans the camera from left to right ending at 0 degrees.
+        resets the camera to 90 degrees.
+        """
+        self.set_pan(90)
+        time.sleep(.2)
+        self.set_pan(0)
+        time.sleep(.2)
+        print ("panning from right to left")
+        for i in range (-20, 200, 2):
```
johnshamoon: Not sure if we should go over/under the limits. Especially if we use numpy.clip. That can be something we can unit test (so we've got more to test later).

This can just be `for i in range (0, 181)`

src/camera.py
```diff
@@ -0,0 +1,212 @@
...
+            self.set_pan(i)
+            time.sleep(.04)
```
johnshamoon: Can we keep all the sleep times the same? also maybe move it into a variable so it's easy to change.

src/camera.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        Tests the panning of the camera
+
+        sets the camera to 90 degrees then to right most value at 0 degrees. 
+        Then pans the camera from right to left giving over the maximum
+        values showing validation function working as intended, ending at 180 
+        degrees. Then pans the camera from left to right ending at 0 degrees.
+        resets the camera to 90 degrees.
+        """
+        self.set_pan(90)
```
johnshamoon: Shouldn't you grab the offset and use it here?

src/camera.py
```diff
@@ -0,0 +1,212 @@
...
+        time.sleep(.2)
+        self.set_pan(0)
+        time.sleep(.2)
+        print ("panning from right to left")
```
johnshamoon: No parenthesis for print (matches the rest of our code).

Also start with a capital and end with a period. Here and below.

src/camera.py
```diff
@@ -0,0 +1,212 @@
+"""
+camera
```
johnshamoon: Since you're adding this module documentation, you should also go enable it in the Sphinx files. You can't generate the docs without the car, but it's really simple to add it. Take a look at the other modules (rst files I believe).

src/camera.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def set_pan(self, degree):
+        """
+        Pans the camera.
+
+        Controls the horizontal movement of the camera in [0, 180] degrees. 
+        0 refers to the vehicle's right while 180 refers to the vehicle's left. 
+
+        :param degree: The degree to set the pan servo to.
+        :type degree: float
```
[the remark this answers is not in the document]
StevenDropiewski: I assume via computer vision we will be getting a lot of no integer values. I tested it with floats and it does use them and works correctly.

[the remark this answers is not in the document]
johnshamoon: yeah but I'm pretty sure all it does is truncate the float part (the hardware isn't that good). It's expected that they'll use ints as the hardware doesn't support panning to say angle 1.2345. It can be used that way, but we shouldn't say that in our documentation.

src/camera.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        self._current_pan_angle = self.validate_degree(degree, 
+                                                       self.MIN_DEGREE,
+                                                       self.MAX_PAN_DEGREE)
+        self._pan_servo.write(self.validate_degree(self._current_pan_angle +
```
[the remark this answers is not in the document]
StevenDropiewski: We can but it may hand the servo class an out of bounds value which it does handle, when their lock value is set to true which it is by default.

src/camera.py
StevenDropiewski: because of the distance the camera has to move it needs a little more leeway for the initial turn rather than the looped panning. I could rename them to long_sleep and short_sleep, These values will only be used in the testing.

src/camera.py
```diff
@@ -3,209 +3,182 @@
...
     cam = Camera()
-    cam.test_pan()
+
+    cam.pan(90)
+    time.sleep(long_interval)
+    cam.pan(0)
+    time.sleep(long_interval)
+    # Tests camera pan.
+    print "Panning from right to left."
+    for i in range (cam.MIN_PAN_DEGREE, cam.MAX_PAN_DEGREE, 2):
```
johnshamoon: No space between range and the parameters. Range is a function and that's weird to do. Also range isn't inclusive on the end condition (i.e. [start, stop) ). It doesn't make a huge difference here but for completeness and as an example, we should fix it.

src/camera.py
```diff
@@ -3,209 +3,182 @@
...
+
+    Creates object and pans from 0 to 180, then from 180 to 0. Then tilts from
+    0 to 150, then 150 to 0.
+    """
+    long_interval = .2
+    short_interval = .04
     cam = Camera()
-    cam.test_pan()
+
+    cam.pan(90)
```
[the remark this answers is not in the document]
johnshamoon: If we don't, we can get rid of long_interval / short_interval and just have one descriptively named timer at 0.05

src/camera.py
johnshamoon: Everywhere else you use range too.

src/camera.py
```diff
@@ -3,209 +3,182 @@
...
+        cam.pan(i)
+        time.sleep(short_interval)
+
+    print "Panning from left to right."
+    for i in range (cam.MAX_PAN_DEGREE, cam.MIN_PAN_DEGREE, -2):
+        cam.pan(i)
+        time.sleep(short_interval)
+
+    # Tests camera tilt.
+    print "tilting up"
```
johnshamoon: Format the tilting comments like the panning ones. Consistency is key.

src/camera.py
```diff
@@ -3,209 +3,182 @@
...
+    """The maximum value to be written to pan servo."""
     MAX_TILT_DEGREE = 150
-    """The maximum value to be written to tilt servo (in degrees)."""
+    """The maximum value to be written to tilt servo."""
     MIN_DEGREE = 0
-    """The minimum value to be written to both pan and tilt servo (in degrees)."""
-    MAX_CALI_DEGREE = 180
-    """The maximum value to calibrate pan and tilt (in degrees)."""
+    """The minimum value to be written to both pan and tilt servo."""
+    MAX_CALI_PAN_DEGREE = 180
```
johnshamoon: Why aren't you using these?

src/camera.py
```diff
@@ -3,209 +3,182 @@
...
+    """The maximum value to calibrate pan."""
+    MAX_CALI_TILT_DEGREE = 150
+    """The maximum value to calibrate tilt."""
     MIN_CALI_DEGREE = -180
-    """The minimum value to calibrate pan or tilt (in degrees."""
-    OFFSET_TO_SERVO = 0
-    """The offset value passed to the servo object."""
+    """The minimum value to calibrate pan or tilt."""
```
johnshamoon: Why is this the same value for both panning and tilting? Shouldn't it be -150 for tilting?

src/camera.py
```diff
@@ -3,209 +3,182 @@
...
 
     def __init__(self, bus_number=1):
+        self.OFFSET_TO_SERVO = 0
```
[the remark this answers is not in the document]
johnshamoon: I think we can do the same for pan_channel and tilt_channel since they're also only ever used in init. Probably better if they're parameters with 1 and 2 (respectively) as the defaults.

src/camera.py
```diff
@@ -3,209 +3,182 @@
...
 
-        :param degree: The amount of degrees to add or subtract for calibration.
-        :type degree: float
+        :param degree: The number of degrees.
+        :type degree: int
         """
-        self._pan_offset = self.validate_degree(degree, self.MIN_CALI_DEGREE,
-                                                        self.MAX_CALI_DEGREE)
-        self._DB.set('pan_offset', self._pan_offset)
+        self._pan_offset = int(numpy.clip(degree, self.MIN_CALI_DEGREE,
```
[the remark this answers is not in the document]
StevenDropiewski: If the user inputs a float value it will be saved, the system won't read it correctly when pulling the data. As long as the user obeys the param type there is no problem, I'm not sure what level of problem prevention we are running at.

src/camera.py
```diff
@@ -3,209 +3,182 @@
...
+        self._db.set('pan_offset', self._pan_offset)
 
 
-    def get_pan(self):
+    def get_pan_angle(self):
         """
+        Gets the current pan angle of the camera.
+
         :return: The current pan angle (in degrees).
         :rtype: float
```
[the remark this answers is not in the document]
StevenDropiewski: The pan(x) where x will still be a float when given by the camera vision and that is what is stored in _current_pan_angle.

src/camera.py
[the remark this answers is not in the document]
johnshamoon: Sure

src/camera.py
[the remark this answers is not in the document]
johnshamoon: Like I said, the SunFounder hardware isn't accurate enough anyway. It'll treat it as an int. It's bad practice to say we take floats when 10.1 and 10.9 do the same thing.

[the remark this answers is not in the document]
StevenDropiewski: gotcha.

src/camera.py
```diff
@@ -0,0 +1,181 @@
...
+    MAX_PAN_CALI_DEGREE = 180
+    """The maximum value to calibrate pan."""
+    MAX_TILT_CALI_DEGREE = 150
+    """The maximum value to calibrate tilt."""
+    MIN_PAN_CALI_DEGREE = -180
+    """The minimum value to calibrate pan."""
+    MIN_TILT_CALI_DEGREE = -150
+    """The minimum value to calibrate tilt."""
+
+    def __init__(self, bus_number=1, pan_channel = 1, tilt_channel = 0, offset = 0):
```
johnshamoon: Shouldn't tilt channel be 2?

johnshamoon: Offset shouldn't be a parameter. That would imply that the user can change it, which would break this because we depend on having a 0 offset (on SunFounder's side anyway). See the db_file in leader/follower to for reference.

src/camera.py
```diff
@@ -0,0 +1,181 @@
...
+        Controls the horizontal movement of the camera in [0, 180] degrees.
+        0 refers to the vehicle's right while 180 refers to the vehicle's left.
+        The degrees will be clamped to [0, 180].
+
+        :param degree: The number of degrees.
+        :type degree: int
+        """
+        self._current_pan_angle = np.clip(degree, self.MIN_DEGREE,
+                                                  self.MAX_PAN_DEGREE)
+        self._pan_servo.write(self._current_pan_angle + self._pan_offset)
```
johnshamoon: Shouldn't this be in the clip validation above? See this example:
```python
self.pan(180)
# self._current_pan_angle is 180
# self._pan_offset > 0, say 10
# self._pan_servo.write(190)
```

johnshamoon: Also if
```python
self.pan(0)
```
with a negative pan offset.

src/camera.py
```diff
@@ -0,0 +1,181 @@
...
+        Controls the vertical movement of the camera in [0, 150] degrees.
+        0 is directly in front of the car, 90 faces upwards, and 150 faces the
+        Raspberry Pi. The degrees will be clamped to [0, 150].
+
+        :param degree: The number of degrees.
+        :type degree: int
+        """
+        self._current_tilt_angle = np.clip(degree, self.MIN_DEGREE,
+                                                   self.MAX_TILT_DEGREE)
+        self._tilt_servo.write(self._current_tilt_angle + self._tilt_offset)
```
johnshamoon: Same issue as panning here.

src/camera.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def main():
+    """
+    Demonstrates the camera's panning and tilting functionality.
+
+    Creates a camera object and pans from 0 to 180, then from 180 to 0. The
+    camera then tilts from 0 to 150, then 150 to 0.
+    """
+    sleep_time = 0.05
```
johnshamoon: The variable name should describe why we need to sleep. It's clear that it's a sleep time, but why is it necessary?

src/camera.py
```diff
@@ -0,0 +1,181 @@
...
+    cam = Camera()
```
johnshamoon: nit:

While we're making changes anyway, can we name this `camera` instead of `cam` ? It just matches the rest of the repo.

src/follower.py
```diff
@@ -137,8 +132,8 @@ def reset_camera(self):
 
         The default position is tilted to 120 degrees and panned to 90 degrees.
         """
-        self._camera.turn_down(120)
-        self._camera.pan_servo.write(90)
+        self._camera.tilt(0)
```
johnshamoon: Change the comment above this too. We're not tilting to 120 degrees here, are we?

src/camera.py
StevenDropiewski: That's why I had it validate twice. Once when setting current pan angle and then before writing to the servo. Otherwise if you give it 90, but you have changed the calibration so 90 is actually 85. _current_pan_angle will then show 85, when it's actually at 90, or what you want to refer to as 90.

src/camera.py
```diff
@@ -75,7 +72,10 @@ def pan(self, degree):
         """
         self._current_pan_angle = np.clip(degree, self.MIN_DEGREE,
                                                   self.MAX_PAN_DEGREE)
-        self._pan_servo.write(self._current_pan_angle + self._pan_offset)
+        self._pan_servo.write(np.clip(self._current_pan_angle +
+                                      self._pan_offset,
+                                      self.MIN_DEGREE,
+                                      self.MAX_DEGREE))
```
johnshamoon: oops

src/camera.py
```diff
@@ -0,0 +1,184 @@
...
+    Creates a camera object and pans from 0 to 180, then from 180 to 0. The
+    camera then tilts from 0 to 150, then 150 to 0.
+    """
+    camera_smoothing = 0.05
+    camera = Camera()
+
+    # Tests camera pan.
+    print "Panning left."
+    for i in range(0, 181, 2):
+        cam.pan(i)
```
johnshamoon: you still have to update the variable name everywhere.

src/camera.py
```diff
@@ -0,0 +1,184 @@
...
+        Pans the camera.
+
+        Controls the horizontal movement of the camera in [0, 180] degrees.
+        0 refers to the vehicle's right while 180 refers to the vehicle's left.
+        The degrees will be clamped to [0, 180].
+
+        :param degree: The number of degrees.
+        :type degree: int
+        """
+        self._current_pan_angle = np.clip(degree, self.MIN_DEGREE,
```
wisambunni: Why isn't _current_pan_angle just the degrees+ offset?

src/camera.py
```diff
@@ -0,0 +1,184 @@
...
+        Tilts the camera.
+
+        Controls the vertical movement of the camera in [0, 150] degrees.
+        0 is directly in front of the car, 90 faces upwards, and 150 faces the
+        Raspberry Pi. The degrees will be clamped to [0, 150].
+
+        :param degree: The number of degrees.
+        :type degree: int
+        """
+        self._current_tilt_angle = np.clip(degree, self.MIN_DEGREE,
```
johnshamoon: why can't we have

```python
self._current_tilt_angle = np.clip(degree + self._tilt_offset, self.MIN_DEGREE, self.MAX_TILT_DEGREE)
self._tilt_servo.write(_current_tilt_angle)
```

wisambunni: Same idea as tilt.

changnor:
Tested API with Steven, system is not additive but the functionality is there with the test function.

johnshamoon:
[Rewrite the SunFounder camera API](https://trello.com/c/DOLPT8rB/89-rewrite-the-sunfounder-camera-api)

johnshamoon:
[Rewrite the SunFounder camera API](https://trello.com/c/DOLPT8rB/89-rewrite-the-sunfounder-camera-api)

johnshamoon:
[Update rest of system to use new camera API](https://trello.com/c/mP5fR3hQ/111-update-rest-of-system-to-use-new-camera-api)

johnshamoon:
[Update rest of system to use new camera API](https://trello.com/c/mP5fR3hQ/111-update-rest-of-system-to-use-new-camera-api)

johnshamoon:
[Update rest of system to use new camera API](https://trello.com/c/mP5fR3hQ/111-update-rest-of-system-to-use-new-camera-api)

johnshamoon:
[Update rest of system to use new camera API](https://trello.com/c/mP5fR3hQ/111-update-rest-of-system-to-use-new-camera-api)

src/camera.py
```diff
@@ -0,0 +1,231 @@
...
+    :type bus_number: int
+
+    :param pan_channel: Pan servo's I2C channel.
+    :type pan_channel: int
+
+    :param tilt_channel: Tilt servo's I2C channel.
+    :type tilt_channel: int
+
+    :param is_cam_absolute: Does the camera work in absolute style.
+    :type cam_style: bool
```
StevenDropiewski: oops.

src/camera.py
```diff
@@ -16,6 +16,7 @@ class Camera:
     Creates a camera object that can be panned and tilted.
 
     The camera can pan in [0, 180] degrees and can tilt in [0, 150] degrees.
+    Operates in absolute degrees default.
```
johnshamoon: complete sentences

src/camera.py
```diff
@@ -42,18 +46,19 @@ class Camera:
     MIN_TILT_CALI_DEGREE = -150
     """The minimum value to calibrate tilt."""
 
-    def __init__(self, bus_number=1, pan_channel = 1, tilt_channel = 2):
-        picar.setup()
-
+    def __init__(self, bus_number=1, pan_channel = 1, tilt_channel = 2,
```
johnshamoon: I'd rather add module level constants

```python
RELATIVE = 0
ABSOLUTE = 1
```
and then name the parameter `pan_type`

johnshamoon: and then you can default to absolute if you want.

```python
pan_type = ABSOLUTE
```

src/camera.py
```diff
@@ -65,17 +70,25 @@ def pan(self, degree):
...
-        The degrees will be clamped to [0, 180].
+        The degrees will be clamped to [0, 180]. Can function in absolute or
+        relative degrees.
 
         :param degree: The number of degrees.
         :type degree: int
         """
-        self._current_pan_angle = np.clip(degree, self.MIN_DEGREE,
-                                                  self.MAX_PAN_DEGREE)
+        if self._is_cam_absolute == True:
```
johnshamoon: and then check 
```python
if self._pan_type == ABSOLUTE:
```

src/camera.py
johnshamoon: And then if a value other than 0 or 1 is passed in, default to absolute

src/camera.py
```diff
@@ -25,6 +26,9 @@ class Camera:
 
     :param tilt_channel: Tilt servo's I2C channel.
     :type tilt_channel: int
+
+    :param is_cam_absolute: Does the camera work in absolute style.
```
johnshamoon: Update these to match the below comments about naming.

src/camera.py
```diff
@@ -65,17 +70,25 @@ def pan(self, degree):
 
         Controls the horizontal movement of the camera in [0, 180] degrees.
         0 refers to the vehicle's right while 180 refers to the vehicle's left.
-        The degrees will be clamped to [0, 180].
+        The degrees will be clamped to [0, 180]. Can function in absolute or
```
johnshamoon: get rid of this change. It's implied in the class docstring. Here and in tilt.

src/camera.py
```diff
@@ -0,0 +1,247 @@
...
+        camera.tilt(i)
+        time.sleep(camera_smoothing)
+
+    # Resetting camera for relative pan.
+    cam.pan(0)
+    time.sleep(camera_smoothing)
+
+    # Remaking object in relative style.
+    camera = None
+    camera = Camera(camera_type = RELATIVE)
```
johnshamoon: You need a `camera.pan(-90)` here

src/leader.py
```diff
@@ -3,13 +3,7 @@
 
 Author: John Shamoon
 """
-from os import path
 import numpy as np
-import sys
-
-FILE_PATH = path.dirname(path.realpath(__file__))
```
johnshamoon: Since we got rid of this, you can get rid of line 18 in docs/conf.py

src/camera.py
wisambunni: `Camera(camera_type=RELATIVE)`

src/camera.py
```diff
@@ -0,0 +1,247 @@
...
+
+    print "Testing relative pan left."
+    for i in range(0, 90, 1):
```
wisambunni: Can be reduced to `range(90)`

src/camera.py
```diff
@@ -0,0 +1,247 @@
...
+        camera.pan(2)
+        time.sleep(camera_smoothing)
+
+    print "Testing relative pan right."
+    for i in range(0, 90, 1):
```
wisambunni: Can be reduced to `range(90)`

src/camera.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def main():
+    """
+    Demonstrates the camera's panning and tilting functionality.
+
+    Creates a camera object and pans from 0 to 180, then from 180 to 0. The
+    camera then tilts from 0 to 150, then 150 to 0.
+    """
+    camera_smoothing = 0.05
```
wisambunni: Should use a better name. This isn't being used to "smooth," just adds a delay in the for loops.

src/camera.py
```diff
@@ -0,0 +1,247 @@
...
+    for i in range(150, -1, -2):
+        camera.tilt(i)
+        time.sleep(camera_smoothing)
+
+    # Resetting camera for relative pan.
+    cam.pan(0)
+    time.sleep(camera_smoothing)
+
+    # Remaking object in relative style.
+    camera = None
```
wisambunni: No need to set the camera to None. Garbage collection takes care of that when you assign it to a new object.

src/camera.py
```diff
@@ -0,0 +1,247 @@
...
+        self._db = filedb.fileDB("config")
+        self._pan_offset = int(self._db.get('pan_offset', default_value=0))
+        self._tilt_offset = int(self._db.get('tilt_offset', default_value=0))
+
+        self._pan_servo = Servo.Servo(pan_channel, bus_number=bus_number,
+                                                   offset=offset)
+        self._tilt_servo = Servo.Servo(tilt_channel, bus_number=bus_number,
+                                                     offset=offset)
+
+        self._camera_type = ABSOLUTE if camera_type is not RELATIVE else RELATIVE
```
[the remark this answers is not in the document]
StevenDropiewski: the camera_type would default to absolute because it did not receive relative.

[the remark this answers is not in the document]
StevenDropiewski: to prevent any accidental inputs. anything besides inputting relative is changed to absolute making it the default.

src/camera.py
```diff
@@ -0,0 +1,266 @@
...
+        camera.pan(2)
+        time.sleep(camera_delay)
+
+    print "Panning right."
+    for i in range(90):
+        camera.pan(-2)
+        time.sleep(camera_delay)
+
+    # Return camera to the center.
+    camera.pan(-90)
```
johnshamoon: Shouldn't this be just 90? If we panned from 0 to 180, then 180 to 0, -90 will keep it where it is.

src/camera.py
```diff
@@ -0,0 +1,266 @@
...
+    time.sleep(camera_delay)
+
+    # Tests realtive camera tilting.
```
johnshamoon: remove this comment

src/camera.py
```diff
@@ -0,0 +1,264 @@
...
+        self._tilt_offset = int(self._db.get('tilt_offset', default_value=0))
+
+        self._pan_servo = Servo.Servo(pan_channel, bus_number=bus_number,
+                                                   offset=offset)
+        self._tilt_servo = Servo.Servo(tilt_channel, bus_number=bus_number,
+                                                     offset=offset)
+
+        self._camera_type = RELATIVE if camera_type is RELATIVE else ABSOLUTE
+
+        self._current_pan_angle = 90
```
johnshamoon: Can you add a function that resets the values to this as well?

i.e. Camera.reset_camera()

src/camera.py
```diff
@@ -0,0 +1,264 @@
...
+        camera.tilt(i)
+        time.sleep(camera_delay)
+
+    print "Tilting down."
+    for i in range(150, -1, -2):
+        camera.tilt(i)
+        time.sleep(camera_delay)
+
+    # Resetting camera for relative panning and tilting.
+    camera.pan(0)
```
johnshamoon: This gets reset when you create a new object anyway, so you can get rid of this.

src/camera.py
```diff
@@ -0,0 +1,271 @@
...
+        return self._current_tilt_angle
+
+
+    def reset_camera(self):
+        """
+        Resets the camera to default position.
+
+        Sets the camera pan to 90 and the camera tilt to 0.
+        """
+        self.pan(90)
```
johnshamoon: This doesn't work for RELATIVE.

docs/index.rst
```diff
@@ -9,6 +9,7 @@ vehicle. Below are the five main modules of the system.
    :maxdepth: 2
 
    bluetoothctl
+   camera
```
johnshamoon: Right above this there's a sentence that says "five main modules." Can you update that to six

src/camera.py
StevenDropiewski: whoops.

src/camera.py
```diff
@@ -0,0 +1,273 @@
...
+        time.sleep(camera_delay)
+
+    print "Tilting down."
+    for i in range(150, -1, -2):
+        camera.tilt(i)
+        time.sleep(camera_delay)
+
+    # Creates a Camera object with relative degrees.
+    camera = Camera(camera_type=RELATIVE)
+    print "Testing relative camera degrees."
```
johnshamoon: Can you move this right above
```python
print "Panning left."
```

for consistency with the above function? Can you just make sure that this everywhere you're doing similar things, they're identical when they can be?

StevenDropiewski: will do.

src/camera.py
```diff
@@ -0,0 +1,276 @@
+"""
+camera
+
+Author: Steven Dropiewski
+"""
+import numpy as np
+from time import sleep
```
johnshamoon: Style: from imports should come before normal imports (sorted alphabetically)

src/camera.py
```diff
@@ -0,0 +1,276 @@
...
+"""The value set to identify relative degrees."""
+
+
+class Camera:
+    """
+    Creates a camera object that can be panned and tilted.
+
+    The camera can pan in [0, 180] degrees and can tilt in [0, 150] degrees.
+    The camera operates in absolute degrees by default, but can be specified to
+    work in realtive degrees during object creation.
```
johnshamoon: realtive

src/camera.py
```diff
@@ -0,0 +1,276 @@
...
+    camera_delay = 0.05
+    picar.setup()
+
+    # Creates a Camera object with absolute degrees (default).
+    camera = Camera()
+    print "Testing absolute camera degrees."
+
+    # Moving the camera to the far right for panning.
+    camera.pan(0)
+
```
johnshamoon: consistency: you should sleep after this line because you sleep on line 230.

StevenDropiewski: agreed. whoops.

src/camera.py
```diff
@@ -186,78 +186,102 @@ def get_tilt_angle(self):
...
 def main():
     """
     Demonstrates the camera's panning and tilting functionality.
 
-    Creates a camera object and pans from 0 to 180, then from 180 to 0. The
-    camera then tilts from 0 to 150, then 150 to 0.
+    Creates a camera object using absolute degrees and pans from 0 to 180, then
+    from 180 to 0. The camera then tilts from 0 to 150, then 150 to 0.
+    Then creates a camera object using relavtive degrees and pans from 0 to 180,
+    then from 180 to 0. The camera then tilts from 0 to 150, then 150 to 0.
```
changnor: Perhaps include delay time units in docstrings.

src/camera.py
```diff
@@ -186,78 +186,102 @@ def get_tilt_angle(self):
...
     picar.setup()
 
     # Creates a Camera object with absolute degrees (default).
     camera = Camera()
-
     print "Testing absolute camera degrees."
+
+    # Moving the camera to the right for panning.
+    camera.pan(0)
+    sleep(camera_delay_long)
```
changnor: Is this where you were attempting to track time or something?

src/camera.py
```diff
@@ -186,78 +186,102 @@ def get_tilt_angle(self):
...
     print "Panning right."
     for i in range(90):
         camera.pan(-2)
-        time.sleep(camera_delay)
+        sleep(camera_delay)
 
-    # Return camera to the center.
-    camera.pan(90)
-    time.sleep(camera_delay)
+    # Moving the camera to the center for tilting.
```
changnor: Moves the camera to center for tilting

src/camera.py
```diff
@@ -186,78 +186,102 @@ def get_tilt_angle(self):
...
     for i in range(75):
         camera.tilt(2)
-        time.sleep(camera_delay)
+        sleep(camera_delay)
 
     print "Tilting down."
     for i in range(75):
         camera.tilt(-2)
-        time.sleep(camera_delay)
+        sleep(camera_delay)
```
changnor: Depending on what you need to do, maybe we can find a different solution instead of the individual sleeps

src/camera.py
StevenDropiewski: No, when the camera requires larger movement the camera_delay time doesn't allow it enough time to travel the full distance

src/camera.py
StevenDropiewski: Without sleeps the physical hardware cannot keep up with python, if both delay variables were set to 0 the main functions would look incredibly different and would be unable to display the abilities of the camera class.

PR #39, opened 2019-03-13
added is_connected method to main.py

StevenDropiewski:


src/main.py
```diff
@@ -1,9 +1,21 @@
...
 
+'''
+Calls hcitool and reads the exit value returned. hcitool returns 0 on success 
+and 1 on failure.
+
+Input: Bluetooth address.
+Output: Boolean values, true if connected false if disconnected.
+'''
+def is_connected(bt_addr):
+    return not subprocess.call(['hcitool','info', bt_addr])
```
johnshamoon: I have a lot of problems with this, but before that:

Can't we just check if /dev/input/js0 exists?

src/main.py
```diff
@@ -1,9 +1,21 @@
...
+
 
-"""
+'''
```
johnshamoon: Don't change these from double to single quotes. It doesn't matter which we use really, but we should be consistent in a file. That means you use the ones that are already there. Don't change the rest of them,

src/main.py
```diff
@@ -1,7 +1,15 @@
+import os
+
 from follower import Follower
 from leader import Leader
 from tagrec import TagRecognition
 
+"""
+Checks if a path exists to /dev/input/js0 and returns a boolean value.
```
johnshamoon: Why does this path matter to someone who doesn't know the system?
`"Returns whether or not a Bluetooth joystick is connected."`

PR #46, opened 2019-03-20
Tune Follower

johnshamoon:
Add a max lost tag time of 0.5 Seconds and gradually increase speed to
smooth movements.


johnshamoon:
[Smooth out follower movements](https://trello.com/c/X1G0GyBm/98-smooth-out-follower-movements)

PR #55, opened 2019-03-31
Changed set speed function to incorporate different accelerations

zeinyoussef:


src/leader.py
```diff
@@ -58,10 +58,12 @@ def set_speed(self, position):
         :param position: The value of the button input.
         :type position: int
         """
+        percent_top_speed = (position + 1)/2
```
johnshamoon: Can we just get rid of the variable for this one? I don't see us tweaking it.

src/leader.py
```diff
@@ -58,10 +58,12 @@ def set_speed(self, position):
...
+
         if position == -1:
             speed = 0
         else:
-            speed = self.MAX_SPEED
+            speed = self.MAX_SPEED * percent_top_speed
```
johnshamoon: Get rid of the "self" here. That's my fault. Somehow slipped through.

johnshamoon: You need to cast this to an int because the SunFounder code freaks out if it's a float. But basically I want this line to be 
```python
speed = int(MAX_SPEED * ((position + 1) / 2))
```

johnshamoon:
Can you also update the middle part of the docstring to be something like
`If the position is -1, the speed is 0. If the position is between (-1, 1), the speed will be between (0, 75). If the position is 1, the speed will be 75.`

changnor:
Math is correct, like John said, self.MAXSPEED just needs to be changed to MAXSPEED as its not a property. Otherwise, it should be good to go.

johnshamoon:
[Variable Acceleration and deceleration to leader](https://trello.com/c/0ctEyG7M/122-variable-acceleration-and-deceleration-to-leader)

PR #62, opened 2019-04-06
Integrated Jevois camera

wisambunni:


src/tagrec.py
```diff
@@ -41,12 +47,19 @@ class TagRecognition():
...
 
-    def __init__(self, resolution=90, dead_zone=1.45, marker_length=0.06):
+    def __init__(self, resolution=90, dead_zone=1.45, marker_length=0.06,
+            contrast=10, brightness=10):
         self._RESOLUTION = resolution
         """The resolution of the camera feed."""
         self._MARKER_LENGTH = marker_length
         """The length of the ARTag in meters."""
 
+        self._CONTRAST = contrast
```
johnshamoon: Should probably verify this value and mention the range of acceptable values in the docstring.

src/tagrec.py
```diff
@@ -41,12 +47,19 @@ class TagRecognition():
...
+        """The camera feed contrast multiplier."""
+
+        self._BRIGHTESS = brightness
```
johnshamoon: Same here and probably also the previous constants

PR #64, opened 2019-04-07
Demo functions

stevendropiewski:


src/leader.py
```diff
@@ -152,6 +163,62 @@ def lead(self):
...
 
 
+    def demo_circle(self):
+        """
+        Controls the leader vehicle in a circle for demo purposes.
+
+        Drives the leader vehicle in a continuous circle till script is
+        terminated.
+        """
+        while True:
```
zeinyoussef: You don't need a while loop here since the entire system is already running one

src/leader.py
```diff
@@ -152,6 +163,62 @@ def lead(self):
...
+
+
+    def demo_line(self):
+        """
+        Controls the leader vehicle in a line for demo purposes.
+
+        Drives the leader vehicle in a line then turns around and repeats till
+        script is terminated.
+        """
+        while True:
```
zeinyoussef: Same here

src/leader.py
```diff
@@ -152,6 +163,62 @@ def lead(self):
...
+
+            straight_timer = time.time()
```
zeinyoussef: You might have to find a way to do this without a timer since it will keep resetting every cycle since you will probably have to take out all of the while loops

src/leader.py
```diff
@@ -152,6 +163,62 @@ def lead(self):
...
+
+    def demo_semicircles(self):
+        """
+        Controls the leader vehicle in semicircles for demo purposes.
+
+        Drives the leader vehicle in a semicircle like path till script is
+        terminated.
+        """
+        while True:
+            right_turn_timer = time.time()
```
zeinyoussef: Same problem here with the while loop and timer

zeinyoussef (changes requested):

src/leader.py
StevenDropiewski: If you called this method without a while loop it would set the front wheels to left turn, set the speed to demo speed, and set back wheels to forward. Then immediately set the car back to the current controller code , position for speed, angle and movement. this allows for testing and some sort of exit statement to be added replacing the while True: statement

johnshamoon: Isn't that supposed to be fixed in the demo mapping though (Zein's branch). This loop never exits this function, which breaks the whole functionality of the system.

src/leader.py
```diff
@@ -152,6 +163,62 @@ def lead(self):
...
+        """
+        Controls the leader vehicle in a line for demo purposes.
+
+        Drives the leader vehicle in a line then turns around and repeats till
+        script is terminated.
+        """
+        while True:
+
+            straight_timer = time.time()
+            while time.time - straight_timer < DEMO_STRAIGHT_TIME:
```
johnshamoon: time.time()

johnshamoon: We don't need this to be in a while loop. Just do those three things, then wait. Otherwise we're sending commands to the wheels over and over again for no reason.

src/leader.py
```diff
@@ -152,6 +163,62 @@ def lead(self):
...
+                self.turn_straight()
+                self.bw.speed(DEMO_SPEED)
+                self.bw.forward()
+
+            turn_timer = time.time()
+            while time.time - turn_timer < DEMO_TURN_TIME:
```
johnshamoon: same down here

src/leader.py
```diff
@@ -31,6 +32,16 @@ class Leader:
 
     STRAIGHT_ANGLE = 90
     """The angle that the hardware associates as straight."""
+    DEMO_SPEED = 40
+    """The speed the leader vehicle drives at."""
+    DEMO_STRAIGHT_TIME = 10
```
johnshamoon: We're going to be using these functions at Design Day. The cars can travel pretty far in 10 seconds. We're not going to have that much space.

src/leader.py
```diff
@@ -31,6 +32,16 @@ class Leader:
...
+    """The time the vehicle drives forward."""
+    DEMO_TURN_TIME = 5
+    """The time the vehicle turns"""
+    DEMO_RIGHT_TURN = 120
```
johnshamoon: Why can't we demo at max turns? Then we can just use the two functions we already have.

src/leader.py
```diff
@@ -152,6 +163,62 @@ def lead(self):
...
+        Drives the leader vehicle in a continuous circle till script is
+        terminated.
+        """
+        while True:
+            self.fw.turn(DEMO_LEFT_TURN)
+            self.bw.speed(DEMO_SPEED)
+            self.bw.forward()
+
+
+    def demo_line(self):
```
johnshamoon: Do we even need this one actually? It just demonstrates an oval basically. The circle is a more intensive demonstration of this. We can show straight line driving ourselves.

src/leader.py
```diff
@@ -31,6 +32,16 @@ class Leader:
 
     STRAIGHT_ANGLE = 90
     """The angle that the hardware associates as straight."""
+    DEMO_SPEED = 40
```
johnshamoon: Why not just use the max speed?

src/leader.py
```diff
@@ -152,6 +163,62 @@ def lead(self):
...
+    def demo_semicircles(self):
+        """
+        Controls the leader vehicle in semicircles for demo purposes.
+
+        Drives the leader vehicle in a semicircle like path till script is
+        terminated.
+        """
+        while True:
+            right_turn_timer = time.time()
+            while time.time - right_turn_timer < DEMO_TURN_TIME:
```
johnshamoon: same here

johnshamoon (changes requested):

johnshamoon (changes requested):

src/leader.py
```diff
@@ -152,6 +157,35 @@ def lead(self):
...
+
+    def demo_semicircles(self):
+        """
+        Controls the leader vehicle in semicircles for demo purposes.
+
+        Drives the leader vehicle in two semicircles.
+        """
+        self.turn_right()
+        self.bw.speed(MAX_SPEED)
+        self.bw.forward()
```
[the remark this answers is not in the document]
StevenDropiewski: right.

src/leader.py
```diff
@@ -152,6 +157,35 @@ def lead(self):
...
+    def demo_circle(self):
+        """
+        Controls the leader vehicle in a circle for demo purposes.
+
+        Sets the wheels and speed of the leader vehicle to drive in a circle.
+        """
+        self.turn_left()
+        self.bw.speed(MAX_SPEED)
+        self.bw.forward()
+        time.sleep(DEMO_CIRCLE_TIME)
```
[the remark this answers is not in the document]
StevenDropiewski: I agree, I originally envisioned it being a sort of listens for button, then enters a while loop so you could use timers instead of sleep, then exits the while loop when you press the button again, or maybe the a button puts the vehicle into demo mode, x and y are demo functions and a to cancel any or all demo functions.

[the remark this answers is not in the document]
johnshamoon: This is my understanding of it. Let's say that:

X -> demo_circle
Y -> demo_semicircles

The user presses the X button. The leader is now doing circles until any other button is pressed. If the button that was pressed was Y, stop doing circles and start doing semi-circles. If the user taps the right trigger, the leader vehicle should get out of the semi_circle mode and go back to waiting for input.

I figured that this functionality would be implemented when mapping the functions and that your functions would just be called repeatedly for as long as the system is in that mode.

zeinyoussef (dismissed):

changnor (dismissed):

src/leader.py
```diff
@@ -31,6 +32,8 @@ class Leader:
 
     STRAIGHT_ANGLE = 90
     """The angle that the hardware associates as straight."""
+    DEMO_TURN_TIME = 2
+    """The time the vehicle takes to turn."""
```
johnshamoon: It does not take the vehicle two seconds to turn. This is how long you turn in demos. This needs a better comment.

src/leader.py
```diff
@@ -152,6 +155,32 @@ def lead(self):
             self.turn(code, position)
 
 
+    def demo_circle(self):
+        """
+        Controls the leader vehicle in a circle for demo purposes.
```
johnshamoon:
```python
def drive_in_circle(self):
    """
    Drives in a circle.
    """
```

src/leader.py
```diff
@@ -152,6 +155,32 @@ def lead(self):
...
+
+        Sets the wheels and speed of the leader vehicle to drive in a circle.
+        """
+        self.turn_left()
+        self.bw.speed(MAX_SPEED)
+        self.bw.forward()
+
+
+    def demo_semicircles(self):
```
johnshamoon:
```python
def drive_wide_winding(self):
    """
    Drives in a wide, snake-like patten.
    """
```

johnshamoon (changes requested):
You're still missing the third demo function. You were supposed to have a smaller snake pattern too.

src/leader.py
```diff
@@ -152,6 +155,41 @@ def lead(self):
...
+    def drives_in_circle(self):
+        """
+        Drives in a circle.
+        """
+        self.turn_left()
+        self.bw.speed(MAX_SPEED)
+        self.bw.forward()
+
+
+    def drive_wide_winding(self):
```
johnshamoon: I think it's cleaner if it's like

```
set speed
forward

turn right
sleep
turn left
sleep
```

src/leader.py
```diff
@@ -152,6 +155,41 @@ def lead(self):
...
+        """
```
johnshamoon: this can be on one line
```"""docstring"""```

Comment (johnshamoon) [reply src/leader.py]: I think we do this in other classes. Double check before doing it.

File: src/leader.py [hunk 19]
```diff
@@ -152,6 +155,41 @@ def lead(self):
...
+        self.bw.speed(MAX_SPEED)
+        self.bw.forward()
+        sleep(DEMO_TURN_TIME)
+
+        self.turn_left()
+        sleep(DEMO_TURN_TIME)
+
+
+    def drive_winding(self):
+        """
```
johnshamoon: Same with this one.

src/leader.py
```diff
@@ -152,6 +155,41 @@ def lead(self):
...
+        Drives in a snake-like pattern.
+        """
+        self.fw.turn(110)
```
johnshamoon: Same comments as above.

src/leader.py
```diff
@@ -152,6 +155,41 @@ def lead(self):
             self.turn(code, position)
 
 
+    def drives_in_circle(self):
+        """
```
johnshamoon: one liner: see below

src/leader.py
```diff
@@ -152,6 +155,41 @@ def lead(self):
             self.turn(code, position)
 
 
+    def drives_in_circle(self):
```
johnshamoon: drive

src/leader.py
```diff
@@ -31,6 +32,8 @@ class Leader:
 
     STRAIGHT_ANGLE = 90
     """The angle that the hardware associates as straight."""
+    DEMO_TURN_TIME = 2
```
johnshamoon: Can we make this 1 second? I feel like that's enough, but we can tune it later if it needs to be 1.5 or 2.

johnshamoon (changes requested):

src/leader.py
StevenDropiewski: ok, will do.

src/leader.py
StevenDropiewski: It is done that way in a few simple functions in the follower, will change in a sec.

src/leader.py
```diff
@@ -152,6 +155,38 @@ def lead(self):
...
+        self.turn_left()
+
+
+    def drive_wide_winding(self):
+        """Drives in a wide, snake-like pattern."""
+        self.bw.speed(MAX_SPEED)
+        self.bw.forward()
+
+        self.turn_right()
+        sleep(DEMO_TURN_TIME)
```
wisambunni: I don't think this works. If you call sleep, then it blocks the whole leader script from doing anything for a whole second until it resumes again to turn right and sleeps some more.

src/leader.py
```diff
@@ -152,6 +155,38 @@ def lead(self):
...
+
+        self.turn_left()
+        sleep(DEMO_TURN_TIME)
```
wisambunni: This won't work.

src/leader.py
```diff
@@ -152,6 +155,38 @@ def lead(self):
...
+
+
+    def drive_winding(self):
+        """Drives in a snake-like pattern."""
+        self.bw.speed(MAX_SPEED)
+        self.bw.forward()
+
+        self.fw.turn(110)
+        sleep(DEMO_TURN_TIME)
```
wisambunni: This won't work.

src/leader.py
```diff
@@ -152,6 +155,38 @@ def lead(self):
...
+
+        self.fw.turn(70)
+        sleep(DEMO_TURN_TIME)
```
wisambunni: This won't work.

src/leader.py
```diff
@@ -152,6 +155,38 @@ def lead(self):
...
+
+
```
wisambunni: We need a clean up function `stop_demo()` for the demos. This function can called when we're going out of demo mode to reset the car to its original state.

wisambunni (changes requested):

johnshamoon:
[Demo functionality](https://trello.com/c/cWH8CIQX/121-demo-functionality)

src/leader.py
```diff
@@ -152,6 +156,76 @@ def lead(self):
...
+        given to the controller return False.
+
+        :return: True if any input is given to the controller.
+        :rtype: Boolean
+
+        :return: False if no input is given to the controller.
+        :rtype: Boolean
+        """
+        code, position = self._controller.get_input()
+        if position is not -1:
```
zeinyoussef: As the button is being released this if statement is going to be entered. You need to be able to check if the button pressed is the one for the demo you are in. I'm not sure how to do this without passing an argument that tells you which demo you are in and ignoring button presses from the button connected to that function

zeinyoussef (changes requested):

src/leader.py
```diff
@@ -152,6 +158,77 @@ def lead(self):
...
+        """
+        Checks if the demo should stop.
+
+        Changes the demo_stop_flag based on any controller input recieved for
+        .1 consecutive seconds. If any button has been pressed for over .1
+        consecutive seconds demo_stop_flag will be set to true. If any button is
+        not being pressed the stop_count is reset.
+        """
+        code, position = self._controller.get_input()
+        sleep(.01)
```
zeinyoussef: This sleep call doesn't do anything. I think there needs to be a better way than sleeping and hoping that the user releases the button in time anyways. If you were to do this, it would have to be before the line where you get controller input but I think it is a really bad solution.

src/leader.py
```diff
@@ -152,6 +158,77 @@ def lead(self):
...
+        if position is 1:
```
zeinyoussef: I think you should do what you were doing before but ignore if the value is -1 or if it is 0. Since -1 means nothing is being pressed and 0 is the state for a released button. Also 0 is very difficult to get to exactly using the triggers and it is the resting position of the 2 joysticks so it would cover almost all cases.

src/leader.py
```diff
@@ -152,6 +158,77 @@ def lead(self):
...
+            self.demo_stop_count += 1
+            if self.demo_stop_count > 10:
```
zeinyoussef: I also wouldn't check for a button to be pressed 10+ times. I'm pretty sure if a button is pressed it only gets recorded once, not once per cycle but testing would need to be done to make sure.

zeinyoussef (changes requested):

src/leader.py
StevenDropiewski: sleep adds a measurable scale to the time being used. Also, only the button being pressed is now being recorded,

src/leader.py
StevenDropiewski: Not pressed 10+ times. demo_stop_count records the amount of times every time code, position = self._controller.get_input() is called when the position value is 1, and every time it calls that method again it grabs a new controller state. the button only needs to be pressed for .1 second to set the flag to break the while loop.

src/leader.py
```diff
@@ -152,6 +158,77 @@ def lead(self):
...
+        self.fw.turn(110)
+        self.demo_turn_wait()
+
+        if self.demo_stop_flag: return
+
+        self.fw.turn(70)
+        self.demo_turn_wait()
+
+
+    def stop_demo_check(self):
```
johnshamoon: I was under the impression that Zein was doing this? I thought it would be a part of the mapping.

src/leader.py
```diff
@@ -152,6 +158,77 @@ def lead(self):
...
+        """
+        Checks if the demo should stop.
+
+        Changes the demo_stop_flag based on any controller input recieved for
+        .1 consecutive seconds. If any button has been pressed for over .1
```
johnshamoon: What? Why would we do this? Why not just end the demo as soon as any input is recorded?

src/leader.py
```diff
@@ -152,6 +158,77 @@ def lead(self):
...
+        consecutive seconds demo_stop_flag will be set to true. If any button is
+        not being pressed the stop_count is reset.
+        """
+        code, position = self._controller.get_input()
```
johnshamoon: Again, I think this should be done in the mapping PR.

src/leader.py
johnshamoon: This sleep does nothing but block and wait. You grab position, wait 0.01 seconds, then check that it is 1. If it is 1 0.01 seconds later, it'll be 1 if you check it immediately.

src/leader.py
johnshamoon: This should be ==. _is_ is for checking if two objects refer to the same object.

johnshamoon: I think this should be done in `lead` with the rest of the mapping, but even still, this doesn't work for anything except buttons.

src/leader.py
johnshamoon: So from what I understand, you're checking to see if they're holding the button for one second?

src/leader.py
```diff
@@ -152,6 +158,77 @@ def lead(self):
...
+        else:
+            self.demo_stop_count = 0
+
+
+    def demo_turn_wait(self):
+        """
+        Waits for the vehicle to turn.
+
+        Waits for the amount of time (in seconds) specified in DEMO_TURN_TIME,
+        is interrupted if demo_stop_flag is made true by stop_demo_check().
```
johnshamoon: too many implementation details.

src/leader.py
```diff
@@ -152,6 +158,77 @@ def lead(self):
...
+    def drive_wide_winding(self):
+        """Drives in a wide, snake-like pattern."""
+        self.demo_stop_flag = False
+        self.bw.speed(MAX_SPEED)
+        self.bw.forward()
+
+        self.turn_right()
+        self.demo_turn_wait()
+
+        if self.demo_stop_flag: return
```
johnshamoon: I think the correct way to do this is with callbacks. What if I want to stop the function right after it starts moving but before it turns?

johnshamoon: also this should be on two lines.

johnshamoon (changes requested):

src/leader.py
StevenDropiewski: As soon as the demo is started it will read a controller input button pressed to start the demo, and most likely read it again, and if any button press terminates the demo it will instantly start and terminate.

src/leader.py
StevenDropiewski: Yeah, I'm really not sure where the line is for what i'm doing and what Zein should be doing for this assignment. I thought it was originally just the commands to send the car in the correct path, but i need some sort of timing mechanism to correctly do the winding functions. Let me know what level you think I should be doing.

src/leader.py
StevenDropiewski: Ok, I've never used callbacks, let me look into it and see what i can do.

PR #68, opened 2019-04-09
Add parameter checking

johnshamoon:


johnshamoon:
[Add bounds checking to set_speed in leader.py](https://trello.com/c/6bx8xEbB/128-add-bounds-checking-to-setspeed-in-leaderpy)

src/leader.py
```diff
@@ -59,10 +59,12 @@ def set_speed(self, position):
         :param position: The value of the button input.
         :type position: int
         """
-        if position == -1:
+        try:
+            position = int(position)
```
zeinyoussef: doesn't this round the value of position to the nearest integer? Wouldn't that mess up the change in speed and only allow for motion in full speed, half speed and stopped?

johnshamoon: Yeah that's my mistake. I was just looking at the docstring. Fixed.

PR #69, opened 2019-04-09
Stop the docs from sorting the contents

johnshamoon:
Methods and constants were sorting alphabetically when they should be sorted by the order that they are in the source.


johnshamoon:
[Stop Sphinx Pages From Sorting Contents Alphabetically](https://trello.com/c/N1TJHqXT/134-stop-sphinx-pages-from-sorting-contents-alphabetically)

PR #74, opened 2019-04-11
Added functionality to bypass camera feed for testing

wisambunni:


src/tagrec.py
```diff
@@ -147,18 +148,30 @@ def make_decision(self, direction):
...
         """
-        self._ret, self._frame = self._cap.read()
+        if not img_src:
+            self._ret, self._frame = self._cap.read()
+        else:
+            file_exists = os.path.isfile(img_src)
+            if file_exists:
+                self._frame = cv2.imread(img_src)
+            else:
+                raise FileNotFoundError('File does not exist')
```
johnshamoon: If we're raising an exception, it needs to be in the docstring.

`:raises FileNotFoundError:`

src/tagrec.py
```diff
@@ -147,18 +148,30 @@ def make_decision(self, direction):
...
 
 
-    def detect(self):
+    def detect(self, img_src=None):
         """
         Detect an ARTag.
 
+        :param img_src: Used to bypass the camera feed and track ARTags from
+                        images instead. Used primarily for testing purposes.
+                        Disabled by default.
```
johnshamoon: Should probably explain that this is a path to an image.

PR #77, opened 2019-04-12
Demo mapping

zeinyoussef:


src/leader.py
```diff
@@ -37,6 +37,7 @@ def __init__(self):
         self._controller = InputController()
 
         db_file = FILE_PATH + "/../" + SUNFOUNDER_PATH + "/config"
+        self._mode = 'normal'
```
johnshamoon: Why are we using strings for modes?

src/leader.py
```diff
@@ -142,14 +143,45 @@ def lead(self):
...
-            self.turn(code, position)
+        if code == 'left_trigger' or code == 'right_trigger':
+            self._mode = 'normal'
+
+        elif code == 'x' and position == 1:
+            if self._mode == 'drives_in_circle':
+                self._mode = 'normal'
+            else:
+                self._mode = 'drives_in_circle'
+
```
johnshamoon: style: remove blank lines before after if statements

src/leader.py
```diff
@@ -142,14 +143,45 @@ def lead(self):
...
+            if code == 'right_trigger':
+                self.set_speed(position)
+                self.drive()
+            if code == 'left_trigger':
+                self.set_speed(position)
+                self.reverse()
+            if code == 'dpad_left_right' or code == 'left_stick_x':
+                self.turn(code, position)
+
+        if self._mode == 'drives_in_circle':
```
johnshamoon: Can't we just do something like:

as long as we don't press another button, call this function. When we press another button, call the function that corresponds to that new button.

src/leader.py
```diff
@@ -142,14 +143,45 @@ def lead(self):
...
         code, position = self._controller.get_input()
-        if code == 'right_trigger':
-            self.set_speed(position)
-            self.drive()
-        if code == 'left_trigger':
-            self.set_speed(position)
-            self.reverse()
-        if code == 'dpad_left_right' or code == 'left_stick_x':
-            self.turn(code, position)
+        if code == 'left_trigger' or code == 'right_trigger':
```
johnshamoon: I think if we press any button, it should jump out of the mode.

johnshamoon (changes requested):

src/leader.py
```diff
@@ -142,14 +143,23 @@ def lead(self):
...
         code, position = self._controller.get_input()
-        if code == 'right_trigger':
-            self.set_speed(position)
-            self.drive()
-        if code == 'left_trigger':
-            self.set_speed(position)
-            self.reverse()
-        if code == 'dpad_left_right' or code == 'left_stick_x':
-            self.turn(code, position)
+        if code is not None:
```
StevenDropiewski: It seems as soon as it calls the function once it will then call lead on the leader object again and return to manual control, we need some sort of way to keep it in a specific mode or calling a specific demo function.

zeinyoussef: If the code is none then it won't switch to manual controls. Also on the release of a button, the code of the button is still called. I tested this and it works the way I think it should.

src/leader.py
zeinyoussef: Strings match the codes for the button's on the controller and allow me to map it easier then other data types since I don't have to write extra lines of code to map a button to an integer or another data type.

johnshamoon: Can't we just say, if we're not in normal mode, run this button's mode (without a series of if elif)? Especially since none of them even take parameters.

src/leader.py
```diff
@@ -142,14 +143,23 @@ def lead(self):
...
+            self._mode = code
```
johnshamoon: style: put a blank line after this. I thought it was a part of the if elif block

johnshamoon (changes requested):

src/leader.py
```diff
@@ -142,14 +148,19 @@ def lead(self):
...
-            self.set_speed(position)
-            self.drive()
-        if code == 'left_trigger':
-            self.set_speed(position)
-            self.reverse()
-        if code == 'dpad_left_right' or code == 'left_stick_x':
-            self.turn(code, position)
+        self._mode = code
+
+        if self._mode in FUNCTION_MAP:
```
johnshamoon:
```python
if code in FUNCTION_MAP
```

This gets rid of the extra variable

src/leader.py
```diff
@@ -142,14 +148,19 @@ def lead(self):
...
+            FUNCTION_MAP[self._mode]
```
johnshamoon: needs self

src/leader.py
```diff
@@ -31,12 +31,18 @@ class Leader:
 
     STRAIGHT_ANGLE = 90
     """The angle that the hardware associates as straight."""
+    FUNCTION_MAP = {
+        'b' : self.drive_winding(),
+        'y' : self.drive_wide_winding(),
+        'x' : self.drives_in_circle() ,
```
johnshamoon: no space between comma and parenthesis.

src/leader.py
johnshamoon: but also, for this to work, don't we need to keep pressing the same button? It should take in the last pressed button and continue to call that until a different input is received.

johnshamoon (changes requested):

PR #80, opened 2019-04-15
Wrote tests for main.py class

zeinyoussef:


test/main/main_test.py
```diff
@@ -0,0 +1,33 @@
...
+
+FILE_PATH = path.dirname(path.realpath(__file__))
+sys.path.append(FILE_PATH + "/../../src")
+
+from leader import Leader
+from follower import Follower
+from tagrec import TagRecognition
+
+class MainTest(unittest.TestCase):
+    def setUp(self):
```
wisambunni: This is not how testing should should be done. You copy pasted the entire functionality of main.py in the test so you're essentially testing the test, not the original code.

PR #82, opened 2019-04-16
Make Main Testable

johnshamoon:


src/main.py
```diff
@@ -10,70 +10,42 @@
...
-    BT_ADDR = "5C:BA:37:26:6D:9A"
+    tag = TagRecognition(marker_length=0.025)
+    vehicle = decide_role(tag.detect())
 
-    # If an ARTag is detected, the vehicle will become a follower.
-    # If an ARTag is not detected, the vehicle will become a leader.
-    if tag.detect():
-        follower = Follower()
+    if isinstance(vehicle, Follower):
         while True:
```
zeinyoussef: Could we get rid of the if else statements and just turn them into while isinstance(vehicle, Follower)? Since the vehicles won't switch roles while running. Or make the while true statement and then have the 2 if statements inside of it?

PR #85, opened 2019-04-17
Camera test

stevendropiewski:


test/camera/camera_test.py
```diff
@@ -0,0 +1,391 @@
...
+import unittest
+
+FILE_PATH = path.dirname(path.realpath(__file__))
+sys.path.append(FILE_PATH + "/../../src")
+
+from camera import Camera
+
+class CameraTest(unittest.TestCase):
+    def setUp(self):
+        self._db = filedb.fileDB(FILE_PATH + "/../camera/config")
```
johnshamoon: this is redundant. just use "config"

test/camera/camera_test.py
```diff
@@ -0,0 +1,391 @@
...
+        self._db.set('pan_offset', 0)
```
johnshamoon: Why not just make these 0 in the config?

test/camera/camera_test.py
```diff
@@ -0,0 +1,391 @@
...
+        camera = Camera()
+        camera.tilt(90)
+        camera.tilt(-1)
+
+        self.assertEqual(camera._current_tilt_angle, 0)
+
+
+    def test_tilt_up_relative(self):
+        camera = Camera(camera_type=RELATIVE)
+        camera.tilt(90)
```
johnshamoon: You don't need to tilt up twice on this one. Do you? can't you just assert that it's at 90?

test/camera/camera_test.py
```diff
@@ -0,0 +1,391 @@
...
+        camera.tilt(30)
+
+        self.assertEqual(camera._current_tilt_angle, 120)
+
+
+    def test_tilt_up_relative_at_bound(self):
+        camera = Camera(camera_type=RELATIVE)
+        camera.tilt(90)
```
johnshamoon: Same here

test/camera/camera_test.py
```diff
@@ -0,0 +1,391 @@
...
+
+        self.assertEqual(camera._current_tilt_angle, 150)
+
+
+    def test_tilt_up_relative_beyond_bound(self):
+        camera = Camera(camera_type=RELATIVE)
+        camera.tilt(90)
+        camera.tilt(61)
+
+        self.assertEqual(camera._current_tilt_angle, 151)
```
johnshamoon: This shouldn't be 151

test/camera/camera_test.py
```diff
@@ -0,0 +1,391 @@
...
+        camera = Camera(camera_type=RELATIVE)
+        camera.tilt(90)
+        camera.tilt(60)
+
+        self.assertEqual(camera._current_tilt_angle, 150)
+
+
+    def test_tilt_up_relative_beyond_bound(self):
+        camera = Camera(camera_type=RELATIVE)
+        camera.tilt(90)
```
johnshamoon: same here

test/camera/camera_test.py
```diff
@@ -0,0 +1,391 @@
...
+
+        self.assertEqual(camera._current_tilt_angle, camera.get_tilt_angle())
+
+
+    def test_calibrate_pan_absolute_increase(self):
+        camera = Camera()
+        camera.calibrate_pan(10)
+
+        self.assertEqual(camera._pan_offset, 10)
```
johnshamoon: You should be testing that it changed in the config file as well.

johnshamoon: for all the below tests as well.

johnshamoon (changes requested):

test/camera/camera_test.py
StevenDropiewski: I need them to be set for the relative calibration tests, if the value is not 0 I can't reliably pinpoint the bound and 1 beyond the bound.

johnshamoon: But you're writing 0 to it anyway. Why not just make it 0 in the config file itself?

test/camera/camera_test.py
```diff
@@ -0,0 +1,448 @@
...
+
+        self.assertEqual(camera._current_tilt_angle, camera.get_tilt_angle())
+
+
+    def test_calibrate_pan_absolute_increase(self):
+        camera = Camera()
+        camera.calibrate_pan(10)
+        offset = int(self._db.get('pan_offset', default_value=0))
+
+        self.assertEqual(camera._pan_offset, 10)
```
johnshamoon: I think you could make this a function.

```python
def calibrate_pan_test(camera, calibration_value, expected_value):
    camera.calibrate_pan(calibration_value)
    offset = int(self._db.get('pan_offset', default_value=0))
    self.assertEqual(camera._pan_offset, expected_value)
    self.assertEqual(offset, expected_value)
```

Now you've reduced every test to two lines instead of 6

johnshamoon (changes requested):
You also want to reset the config file to what it was after every test. Each test should have the same starting point. You should also have a tearDown function that resets everything you changed.

test/camera/camera_test.py
StevenDropiewski: Ah, was not sure if you could do that in testing. Much cleaner.

test/camera/camera_test.py
```diff
@@ -0,0 +1,391 @@
...
+        camera.pan(60)
+        camera.tilt(90)
+
+        camera.reset_camera()
+
+        self.assertEqual(camera._current_pan_angle, 90)
+        self.assertEqual(camera._current_tilt_angle, 0)
+
+
+def calibrate_pan_test(camera, calibration_value, expected_value):
```
johnshamoon: Why aren't these a part of the class?

test/camera/camera_test.py
```diff
@@ -0,0 +1,391 @@
...
+    def test_get_tilt_angle(self):
+        camera = Camera()
+        camera.tilt(90)
+
+        self.assertEqual(camera._current_tilt_angle, camera.get_tilt_angle())
+
+
+    def test_calibrate_pan_absolute_increase(self):
+        camera = Camera()
+
```
johnshamoon: Can you remove the blank space between these in all your tests? I think it looks a lot cleaner for the two line tests.

johnshamoon (changes requested):

test/camera/camera_test.py
StevenDropiewski: can do.

test/camera/camera_test.py
StevenDropiewski: I was not exactly sure the correct way to call in the object and am unable to test the specific structure in unittest without the picar. I got it to work in a sudo test file outside of the class, let me know if there are any errors or what happens if you get time to run it. thanks.

PR #89, opened 2019-04-17
Leader testing

changnor:


test/leader/README.md
```diff
@@ -0,0 +1,4 @@
+# Leader Tests
+## Prerequisites
+1. Leader vehicle must be powered on during all testing. 
+2. Controller must be connected to vehicle for the duration of testing in order to give input.
```
johnshamoon: Controller.

The controller should not be giving any input.

test/leader/leader_test.py
```diff
@@ -0,0 +1,74 @@
...
+class LeaderTest(unittest.TestCase):
+    def test_set_speed_lower_bound(self):
+        leader = Leader()
+
+        self.assertEqual(leader.set_speed(-1), 0)
+
+
+    def test_set_speed_upper_bound(self):
+        leader = Leader()
+
```
johnshamoon: this blank space is unnecessary and makes the tests harder to follow.

test/leader/leader_test.py
```diff
@@ -0,0 +1,74 @@
...
+        self.assertEqual(leader.set_speed(1) , 50)
```
johnshamoon: set speed doesn't return anything.

johnshamoon: don't put a space between the parenthesis and the comma

johnshamoon: And why 50? What if I change the max speed?

test/leader/leader_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+    def test_set_speed_zero(self):
+        leader = Leader()
+
+        self.assertEqual(leader.set_speed(0) , 50/2)
```
johnshamoon: again, why 50?

test/leader/leader_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def test_reverse_zero_speed(self):
+        leader = Leader()
+
+        self.assertEqual(leader.set_speed(-1) , 0)
+
+
+if __name__ == '__main__':
+    unittest.main()
+
```
johnshamoon: fix whatever happened here too

test/leader/leader_test.py
changnor: Should I remove all the blank spaces before the asserts? I followed Wisam's style, should I not?

test/leader/leader_test.py
changnor: I followed the test case specification, the input and desired output was specified.

changnor: instead of 50 i'll just put leader.MAX_SPEED

test/leader/README.md
changnor: modified to driver gives input

test/leader/leader_test.py
changnor: I removed the blank spaces, does that look better?

test/leader/leader_test.py
changnor: Its not complaining anymore

